PR #75, opened 2022-05-09
Fixing wrong xml attribute in README.md

khmaies5:
replacing 
```xml 
app:chartType="column"
```
with

```xml
app:chart="column"
```

using app:chartType="column" will produce an error and in the sample app you use app:chart="column"


patrykmichalik (approved):

patrykmichalik:
Thank you, @khmaies5!

Issue #90, opened 2022-06-02
Weird display when isHorizontalScrollEnabled=false


JamalMulla:
When `isHorizontalScrollEnabled=false`, the chart is displayed as a single vertical line. It only displays properly after the chart is touched. This doesn't happen if the flag isn't set. Is there any other way to show the whole data without scrolling?

I'm using the jetpack compose Chart with lineChart()

patrykmichalik:
Hello! Thanks for the bug report. We can’t seem to reproduce this issue on our end. Would you be able to provide a minimal reproducible example and a screenshot or screen recording of the bug? Cheers!

JamalMulla:
I've attached a video of the bug along with a minimal example that reproduces it in JC. I'm using v1.0.0. In this minimal example, I actually need to try and zoom for the graph to show up.

```
@Composable
fun ComposeLineChart(
    color: Color,
) {

    val stuff = entryModelOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19,)

    ProvideChartStyle(chartStyle = m3ChartStyle(entityColors = listOf(color))) {

        Chart(
            chart = lineChart(),
            model = stuff,
            startAxis = startAxis(),
            bottomAxis = bottomAxis(),
            isHorizontalScrollEnabled = false,
        )
    }
}
```
[Video](https://imgur.com/a/YOZbeXk)

patrykmichalik:
Thanks, @JamalMulla! I’ve managed to reproduce the issue and we’re looking into it. Apologies for the inconvenience.

patrykmichalik:
Hello again! This issue is resolved in Vico 1.1.0, which should be available within half an hour. Thanks for the bug report!

PR #91, opened 2022-06-06
Fix charts being drawn incorrectly when horizontal scrolling is disabled

patrykmichalik:
This fixes issue #90.


vico/core/src/main/java/com/patrykandpatryk/vico/core/layout/VirtualLayout.kt
```diff
@@ -83,11 +86,18 @@ public open class VirtualLayout(
             finalInsets.setValuesIfGreater(tempInsets)
         }
 
+        val chartBounds = RectF().apply {
```
Gowsky, line 89: This property seems useless. All of its values are read only once in `chart.setBounds`.

patrykmichalik: The variable is passed to `AxisManager#setAxesBounds` in line 103. `chartBounds` is no longer a field of `context` due to the type of `context` having been changed.

Gowsky: Alright.

PR #99, opened 2022-07-02
Add initial chart scale settings

margarita-v:
Sometimes it's important to set up the initial chart scale
Снимок экрана 2022-07-02 в 20 01 00

### DIFF
--- a/sample/src/main/java/com/patrykandpatryk/vico/sample/chart/LineChart.kt
+++ b/sample/src/main/java/com/patrykandpatryk/vico/sample/chart/LineChart.kt
@@ -48,6 +48,7 @@ internal fun ComposeLineChart(
             startAxis = startAxis,
             bottomAxis = bottomAxis,
             marker = marker(),
+            chartScale = 3f
         )
     }
 }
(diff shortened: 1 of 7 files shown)


patrykmichalik:
Hello! Cheers for your interest in contributing. Would you be able to expand on the use case here? I believe a result identical to the one in the image could be achieved by increasing the segment width. Does allowing the initial scale to be changed provide any benefit over that?

margarita-v:
@patrykmichalik Could you elaborate how can I increase the segment width (in compose and view chart)? I'll check and maybe it would be enough

patrykmichalik:
Sure. In the case of line charts, you can bump up the point spacing. In Compose, thatʼd look like so: `lineChart(spacing = 64.dp, ...)`. In the view system, you could use this: `(chartView.chart as LineChart).spacingDp = 32f` (where `chartView` is a reference to your `ChartView` instance). For column charts, you can set the column width, the spacing between multiple columns in a single segment, and the spacing between the edges of each segment and the column or columns it contains.

margarita-v:
@patrykmichalik Thank you, it helps :) I think that PR should be closed

patrykmichalik:
You’re welcome, and thanks again!

PR #107, opened 2022-07-19
Cause `vi`ew`co`mpose

AntsyLich:


patrykmichalik:
Thanks for the PR! We generally maintain an order in which Compose comes first and the view system comes second. This is true for both the wiki and the sample app, and since weʼd like to keep things consistent, weʼd rather Compose came first in the README too. Cheers!

PR #114, opened 2022-08-06
Introduce `LineChart.PointHorizontalPosition`, `LineChart.LineSpec.PointConnector`, and `HorizontalAxis.LabelPosition`

patrykmichalik:


vico/core/src/main/java/com/patrykandpatryk/vico/core/axis/horizontal/HorizontalAxis.kt
```diff
@@ -162,6 +170,19 @@ public class HorizontalAxis<Position : AxisPosition.Horizontal>(
     private fun getEntryLength(segmentWidth: Float) =
         ceil(bounds.width() / segmentWidth).toInt() + 1
 
+    private fun forEachEntityIndex(
```
Gowsky, line 173: Making this function `inline` may slightly improve its performance.
```suggestion
    private inline fun forEachEntityIndex(
```

Gowsky (approved):

Issue #117, opened 2022-08-13
Customize or disable chart animation


margarita-v:
Thanks again for the great library :)
I use it in app with fragments, navigation component, bottom bar, render the chart data in `onViewCreated`. But the chart animation is repeated whenever I enter the tab again though the data should be the same. Is it a way to disable the animation or customize it?

https://user-images.githubusercontent.com/22121595/184478833-33280a0f-bbac-4408-a3fb-86c28840c610.mp4

patrykmichalik:
Hello! We’re glad Vico has been useful to you. The next update will introduce the ability to disable the animation in question. I’ll let you know once the update is available.

PR #120, opened 2022-08-14
Add ability to disable initial animation for charts

patrykmichalik:
This resolves issue #117.

### DIFF
--- a/vico/view/src/main/java/com/patrykandpatryk/vico/view/chart/BaseChartView.kt
+++ b/vico/view/src/main/java/com/patrykandpatryk/vico/view/chart/BaseChartView.kt
@@ -162,6 +162,12 @@ public abstract class BaseChartView<Model : ChartEntryModel> internal constructo
     public var isZoomEnabled: Boolean = true
 
     /**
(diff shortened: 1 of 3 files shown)


vico/view/src/main/java/com/patrykandpatryk/vico/view/chart/BaseChartView.kt
```diff
@@ -161,6 +161,12 @@ public abstract class BaseChartView<Model : ChartEntryModel> internal constructo
      */
     public var isZoomEnabled: Boolean = true
 
+    /**
+     * Whether to display an animation when the chart is created. In this animation, the value of each chart entry is
+     * animated from zero to the actual value.
+     */
+    public var initialAnimation: Boolean = true
```
Gowsky, line 168: Wouldn’t it be better to name it e.g. `runInitialAnimation`? `initialAnimation` would suggest that it defines the animation itself.

patrykmichalik: Please see 513bd3097b3c7258712dcf5d45db1144edb04111.

Gowsky (approved):

PR #129, opened 2022-09-10
[Sample] Add ability to click on arrows within SwipeHint

evowizz:
This PR adds the ability to click on arrows within the `SwipeHint` (located below the charts in the sample.) While this isn't the most useful PR of all, it fixes a small confusion I've had multiple times while trying out the sample app.

The ability to swipe between pages remains available and the ability to click the arrows does not break it.

### DIFF
--- a/sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
+++ b/sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
@@ -99,7 +99,7 @@ internal fun ChartPager(
                     currentPage = state.currentValue + 1,
                     pageCount = itemCount,
                 ) { direction ->
-                    val newState = (state.currentValue + direction)
+                    val newState = state.currentValue + direction
                     if (newState in 0 until itemCount) {
                         scope.launch { state.animateTo(newState) }
                     }
--- a/sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
+++ b/sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
@@ -30,6 +30,7 @@ import androidx.compose.material.SwipeableState
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
@@ -37,6 +38,7 @@ import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import com.patrykandpatryk.vico.sample.util.SampleChart
 import com.patrykandpatryk.vico.sample.util.Tab
+import kotlinx.coroutines.launch
 
 @Composable
 @OptIn(ExperimentalMaterialApi::class)
@@ -46,6 +48,8 @@ internal fun ChartPager(
     sampleCharts: List<SampleChart>,
     tab: Tab,
 ) {
+    val scope = rememberCoroutineScope()
+
     HorizontalPager(
         state = state,
         itemCount = itemCount,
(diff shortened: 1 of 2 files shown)


sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
```diff
@@ -93,8 +97,11 @@ internal fun ChartPager(
             ) {
                 SwipeHint(
                     currentPage = state.currentValue + 1,
-                    pageCount = itemCount,
-                )
+                    pageCount = itemCount
```
patrykmichalik, line 100: It seems that a trailing comma is missing here.

```suggestion
                    pageCount = itemCount,
```

sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
```diff
@@ -93,8 +97,11 @@ internal fun ChartPager(
...
+                ) { direction: Int ->
```
patrykmichalik, line 101: For consistency with the rest of the source code, I’d suggest removing the type annotation here.

```suggestion
                ) { direction ->
```

sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
```diff
@@ -93,8 +97,11 @@ internal fun ChartPager(
...
+                    val newState = (state.currentValue + direction).coerceIn(0 until itemCount)
+                    scope.launch { state.animateTo(newState) }
```
patrykmichalik, line 103: I think it would be preferable for `animateTo` not to be called at all when `state.currentValue + direction` is outside of the `0 until itemCount` range. This would reduce the number of operations.

```suggestion
                    scope.launch {
                        (state.currentValue + direction)
                            .takeIf { it in 0 until itemCount }
                            ?.let { state.animateTo(it) }
                    }
```

patrykmichalik (changes requested):
Thanks for the PR! Here are some suggestions.

sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
evowizz: Oh, this isn't required by Kotlin. Besides, all parameters are already provided.

patrykmichalik: While trailing commas are optional, the Kotlin style guide recommends that they be used—please see [here](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas).

evowizz: Fair enough 👍 You may want to add a code style to the project.

sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
```diff
@@ -94,7 +98,12 @@ internal fun ChartPager(
                 SwipeHint(
                     currentPage = state.currentValue + 1,
                     pageCount = itemCount,
-                )
+                ) { direction ->
+                    val newState = (state.currentValue + direction)
```
patrykmichalik, line 102: These parentheses seem to be unnecessary.

```suggestion
                    val newState = state.currentValue + direction
```

patrykmichalik (changes requested):
Thanks! One last suggestion.

patrykmichalik (approved):

patrykmichalik:
Thanks, @evowizz!

sample/src/main/java/com/patrykandpatryk/vico/sample/ui/ChartPager.kt
patrykmichalik: It looks like `ALLOW_TRAILING_COMMA` is set to `true` in the project’s code style (see [here](https://github.com/patrykandpatryk/vico/blob/development/.idea/codeStyles/Project.xml#L15)). I’ve just added two rules connected to trailing commas to the `.editorconfig` file (see [here](https://github.com/patrykandpatryk/vico/blob/development/.editorconfig#L11-L13)).

Issue #157, opened 2022-11-21
Bug: Column thickness and spacing changes depending on the data count


risalfajar:
This is the preview I get from Jetpack Compose, you can see that the more the data displayed, the smaller the column is.

![image](https://user-images.githubusercontent.com/28381124/203094167-35b24c95-2a9d-4653-a036-363eecf4573c.png)

**Expected behavior**
Column thickness and spacing should be consistent regardless of how much data is displayed, if the data is too much to display in the screen, then it should be horizontally scrollable.

Code:
```kotlin
Chart(
    chart = columnChart(
        columns = listOf(
            LineComponent(context.getColor(R.color.primary), 1f, shape = pillShape),
            LineComponent(context.getColor(R.color.secondary), 1f, shape = pillShape),
        ),
        dataLabel = textComponent {
            typeface = Typeface.SANS_SERIF
            textAlign = Paint.Align.CENTER
            color = android.graphics.Color.WHITE
        },
        dataLabelVerticalPosition = VerticalPosition.Bottom,
        spacing = 4.dp,
        innerSpacing = 0.1.dp
    ),
    model = entryModelOf(frequencies, remainingStocks),
    startAxis = startAxis(
        label = axisLabelComponent(typeface = Typeface.SANS_SERIF, textSize = 10.sp, lineCount = 1),
        axis = null,
        tick = null,
        guideline = axisGuidelineComponent(shape = Shapes.rectShape, color = Color.LightGray.copy(alpha = .25f)),
        valueFormatter = DecimalFormatAxisValueFormatter("##")
    ),
    bottomAxis = bottomAxis(
        label = axisLabelComponent(typeface = Typeface.SANS_SERIF, textSize = 10.sp, lineCount = 1),
        guideline = null,
        tick = null,
        axis = null,
        valueFormatter = { value, _ -> labels[value.roundToInt()] }
    ),
)
```

risalfajar:
To add more context, in the app I'm working on, the columns count is dynamic, it isn't fixed like _January-December_ for example. It can be few or many, depending on the data sent by the API. That's why I want to make the column thickness static regardless of columns count.

Yes I'm agreed, I think we need parameter like in point 2.

And if I understand your first point right, shouldn't `thicknessDp` named `minThicknessDp` instead?

patrykmichalik:
[...]

PR #160, opened 2022-11-26
Introduce `AutoScaleUp`

patrykmichalik:
This addresses issue #157.

### DIFF
--- a/vico/compose/src/main/java/com/patrykandpatryk/vico/compose/chart/Charts.kt
+++ b/vico/compose/src/main/java/com/patrykandpatryk/vico/compose/chart/Charts.kt
@@ -56,6 +56,7 @@ import com.patrykandpatryk.vico.core.chart.draw.chartDrawContext
 import com.patrykandpatryk.vico.core.chart.draw.drawMarker
 import com.patrykandpatryk.vico.core.chart.draw.getMaxScrollDistance
 import com.patrykandpatryk.vico.core.chart.edges.FadingEdges
+import com.patrykandpatryk.vico.core.chart.scale.AutoScaleUp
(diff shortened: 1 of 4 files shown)


vico/compose/src/main/java/com/patrykandpatryk/vico/compose/chart/Charts.kt
```diff
@@ -172,6 +182,8 @@ public fun <Model : ChartEntryModel> Chart(
     legend: Legend? = null,
     isHorizontalScrollEnabled: Boolean,
     isZoomEnabled: Boolean = true,
+    fadingEdges: FadingEdges? = null,
+    autoScaleUp: AutoScaleUp = AutoScaleUp.Full,
```
Gowsky, line 186: We don't need to add these arguments to a deprecated function, as we want to remove it sooner or later. If someone wants use new features, they will have to use one of non-deprecated functions.

Gowsky (changes requested):

vico/compose/src/main/java/com/patrykandpatryk/vico/compose/chart/Charts.kt
patrykmichalik: I agree that this isn’t necessary, but doesn’t it reduce the probability of confusion arising? We’re currently using `DeprecationLevel.Warning`, meaning the function remains available, but we recommend switching to the non-deprecated counterpart. Still, if someone is using the deprecated function, which they can, it won’t be immediately clear that `fadingEdges` and `autoScaleUp` exist. I think we could keep these parameters and increment the deprecation level to `DeprecationLevel.Error` in the near future in order to have people switch to the non-deprecated function.

Gowsky (approved):

Issue #177, opened 2023-01-22
NullPointerException on MutableMeasureContext.setLtr(boolean)


KesWalker:
Hi, I run into this exception when I try and open a fragment with a chart in. 

```
   Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.patrykandpatryk.vico.core.context.MutableMeasureContext.setLtr(boolean)' on a null object reference
        at com.patrykandpatryk.vico.view.chart.BaseChartView.onRtlPropertiesChanged(BaseChartView.kt:436)
        at android.view.View.resolvePadding(View.java:21086)
        at android.view.View.initializeScrollbarsInternal(View.java:6838)
        at android.view.View.<init>(View.java:6168)
        at android.view.View.<init>(View.java:5474)
        at com.patrykandpatryk.vico.view.chart.BaseChartView.<init>(BaseChartView.kt:79)
E/AndroidRuntime:     at com.patrykandpatryk.vico.view.chart.ChartView.<init>(ChartView.kt:32)
        at com.patrykandpatryk.vico.view.chart.ChartView.<init>(ChartView.kt:28)
        at com.patrykandpatryk.vico.view.chart.ChartView.<init>(Unknown Source:11)
        	... 38 more
```

This only happens periodically for some reason, not everytime. Most times it works fine. I notice it seems to be happening on right to left property changes, I am not making any config changes that I'm aware of so I'm not sure why that's happening.

Gowsky:
Hi @KesWalker! Thanks for the report. We’ll take a look at it.

KesWalker:
Thanks @Gowsky , this is the leading crash in my app so I look forward to seeing it resolved without having to remove the charts :)

Gowsky:
@KesWalker 
>this is the leading crash in my app 

>I notice it seems to be happening on right to left property changes,

Do users often switch between LTR and RTL layouts?
Do you know if this crash affects all android versions, or only a particular one?

KesWalker:
This is happening without anyone changing text config from LTR/RTL, it's happening primarily on Android 13.On 26 Jan 2023 16:28, Patryk Goworowski ***@***.***> wrote:
@KesWalker

this is the leading crash in my app


I notice it seems to be happening on right to left property changes,

Do users often switch between LTR and RTL layouts?
Do you know if this crash affects all android versions, or only a particular one?

patrickmichalik:
[...]

PR #179, opened 2023-01-27
Fix NPE in `BaseChartView#onRtlPropertiesChanged` #177

Gowsky:


vico/views/src/main/java/com/patrykandpatrick/vico/views/chart/BaseChartView.kt
```diff
@@ -500,7 +500,13 @@ public abstract class BaseChartView<Model : ChartEntryModel> internal constructo
...
 
+    @Suppress("SENSELESS_COMPARISON")
     override fun onRtlPropertiesChanged(layoutDirection: Int) {
-        measureContext.isLtr = layoutDirection == ViewCompat.LAYOUT_DIRECTION_LTR
+        // This function may be invoked inside of the View’s constructor, before the measureContext is initialized.
+        // In this case, we can ignore this callback, as the layout direction will be determined when the MeasureContext
+        // instance is created.
+        if (measureContext != null) {
+            measureContext.isLtr = layoutDirection == ViewCompat.LAYOUT_DIRECTION_LTR
+        }
```
patrickmichalik, line 510: I’d suggest using a guard clause here.

```suggestion
        if (measureContext == null) return
        measureContext.isLtr = layoutDirection == ViewCompat.LAYOUT_DIRECTION_LTR
```

patrickmichalik (changes requested):

patrickmichalik (approved):

Issue #200, opened 2023-03-08
Crash when graph get's resized to zero height


MFlisar:
I Just placed a graph inside my layout and my layout is collapsible and resizes its content. If I place your graph inside it it crashes (when it gets resized to 0, because it tries to call `coerceIn` with an empty or range).

It's probably a rare use case, but the graph should at least not crash if it gets resized to a zero height...  My suggestion is to use 0 in such a case instead of `coerceIn` which in case of the exception I see below would be the correct y value if the graph is to small to draw anything on the y axis anyway... Or alternatively skip the drawing at all...

**Version:**
compose-m3:1.6.4

**Code**
```
val chartEntryModel = entryModelOf(4f, 12f, 8f, 16f, 20f, 5f, 34f, 72f, 15f, 4f, 12f, 8f, 16f, 20f, 5f, 34f, 72f, 15f)
Chart(
	modifier = Modifier
		.fillMaxWidth()
		.height(100.dp)
		.padding(8.dp),
	chart = lineChart(),
	model = chartEntryModel,
	startAxis = startAxis(),
	bottomAxis = bottomAxis()
)
```

**Stacktrace**
```
java.lang.IllegalArgumentException: Cannot coerce value to an empty range: maximum -30.5 is less than minimum 27.5.
	at kotlin.ranges.RangesKt___RangesKt.coerceIn(_Ranges.kt:1441)
	at com.patrykandpatrick.vico.core.chart.line.LineChart$drawChart$1$1$1.invoke(LineChart.kt:340)
	at com.patrykandpatrick.vico.core.chart.line.LineChart$drawChart$1$1$1.invoke(LineChart.kt:301)
	at com.patrykandpatrick.vico.core.chart.line.LineChart.forEachPointWithinBoundsIndexed(LineChart.kt:469)
	at com.patrykandpatrick.vico.core.chart.line.LineChart.drawChart(LineChart.kt:301)
	at com.patrykandpatrick.vico.core.chart.BaseChart.drawChartInternal(BaseChart.kt:133)
	at com.patrykandpatrick.vico.core.chart.BaseChart.drawScrollableContent(BaseChart.kt:92)
	at com.patrykandpatrick.vico.core.chart.BaseChart.drawScrollableContent(BaseChart.kt:37)
	at com.patrykandpatrick.vico.compose.chart.ChartsKt$ChartImpl$6.invoke(Charts.kt:382)
[...]

PR #203, opened 2023-03-11
Fix a crash occurring when chart bounds are invalid #200

Gowsky:


vico/core/src/main/java/com/patrykandpatrick/vico/core/extension/RectExtensions.kt
```diff
@@ -21,6 +21,12 @@ import kotlin.math.abs
 import kotlin.math.cos
 import kotlin.math.sin
 
+/**
+ * Defines whether bounds within this [RectF] are invalid and can’t be used to draw a content.
+ */
+public val RectF.hasInvalidBounds: Boolean
+    get() = top >= bottom || left >= right
```
patrickmichalik, line 28: This looks like a duplicate of `RectF#isEmpty`.

vico/core/src/main/java/com/patrykandpatrick/vico/core/layout/VirtualLayout.kt
```diff
@@ -52,6 +52,8 @@ public open class VirtualLayout(
      * @param legend the legend for the chart.
      * @param segmentProperties the [SegmentProperties] of the chart.
      * @param chartInsetter additional components that influence the chart layout, such as markers.
+     *
+     * @return the bounds set to the [chart].
```
patrickmichalik, line 56:
```suggestion
     * @return the bounds applied to the chart.
```

patrickmichalik (changes requested):

vico/core/src/main/java/com/patrykandpatrick/vico/core/extension/RectExtensions.kt
Gowsky: You’re right!

patrickmichalik (approved):

PR #260, opened 2023-05-14
Add Horizontal Legend Implementation.

lx200916:
`HorizontalLegend` Displays items in a line wrapping horizontal row, so that it may save more vertical space.😊

### DIFF
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/legend/Legends.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/legend/Legends.kt
@@ -32,10 +32,10 @@ import com.patrykandpatrick.vico.core.legend.VerticalLegend
 /**
  * Creates a [VerticalLegend].
  *
- * @param items a [Collection] of [Item]s to be displayed by this [VerticalLegend].
- * @param iconSize defines the size of all [Item.icon]s.
- * @param iconPadding defines the padding between each [Item.icon] and its corresponding [Item.label].
- * @param spacing defines the vertical spacing between each [Item].
+ * @param items a [Collection] of [LegendItem]s to be displayed by this [VerticalLegend].
+ * @param iconSize defines the size of all [LegendItem.icon]s.
+ * @param iconPadding defines the padding between each [LegendItem.icon] and its corresponding [LegendItem.label].
+ * @param spacing defines the vertical spacing between each [LegendItem].
  * @param padding defines the padding of the content.
  */
 @Composable
@@ -63,7 +63,7 @@ public fun verticalLegend(
  * @param labelText the text content of the label.
  */
 @Composable
-public fun LegendItem(
+public fun legendItem(
     icon: Component,
     label: TextComponent,
     labelText: CharSequence,
@@ -78,11 +78,11 @@ public fun LegendItem(
 /**
  * Creates a [HorizontalLegend].
  *
- * @param items a [Collection] of [Item]s to be displayed by this [HorizontalLegend].
- * @param iconSize defines the size of all [Item.icon]s.
- * @param iconPadding defines the padding between each [Item.icon] and its corresponding [Item.label].
- * @param lineSpacing defines the vertical spacing between each line of [Item]s.
- * @param spacing defines the vertical spacing between each [Item].
+ * @param items a [Collection] of [LegendItem]s to be displayed by this [HorizontalLegend].
+ * @param iconSize defines the size of all [LegendItem.icon]s.
+ * @param iconPadding defines the padding between each [LegendItem.icon] and its corresponding [LegendItem.label].
+ * @param lineSpacing defines the vertical spacing between each line of [LegendItem]s.
+ * @param spacing defines the vertical spacing between each [LegendItem].
  * @param padding defines the padding of the content.
  */
 
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/legend/Legends.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/legend/Legends.kt
@@ -56,6 +56,34 @@ public fun verticalLegend(
 }
 
 /**
+ * Creates a [VerticalLegend].
+ *
+ * @param items a [Collection] of [VerticalLegend.Item]s to be displayed by this [VerticalLegend].
+ * @param iconSize defines the size of all [LegendItem.icon]s.
+ * @param iconPadding defines the padding between each [LegendItem.icon] and its corresponding [LegendItem.label].
+ * @param spacing defines the vertical spacing between each [LegendItem].
+ * @param padding defines the padding of the content.
+ */
+@JvmName("verticalLegendWithItem")
+@Composable
+@Deprecated("verticalLegendItem and VerticalLegend.Item is deprecated. Use legendItem and LegendItem instead.")
+public fun verticalLegend(
+    items: Collection<VerticalLegend.Item>,
+    iconSize: Dp,
+    iconPadding: Dp,
+    spacing: Dp = 0.dp,
+    padding: MutableDimensions = emptyDimensions(),
+): VerticalLegend = remember(items, iconSize, iconPadding, spacing, padding) {
+    VerticalLegend(
+        items = items.map { LegendItem(it.icon, it.label, it.labelText) },
+        iconSizeDp = iconSize.value,
+        iconPaddingDp = iconPadding.value,
+        spacingDp = spacing.value,
+        padding = padding,
+    )
+}
+
+/**
  * Defines the appearance of an item of a [Legend].
  *
  * @param icon the [Component] used as the item’s icon.
@@ -64,6 +92,27 @@ public fun verticalLegend(
  */
 @Composable
 public fun legendItem(
+    icon: Component,
+    label: TextComponent,
+    labelText: CharSequence,
+): LegendItem = remember(icon, label, labelText) {
+    LegendItem(
+        icon = icon,
+        label = label,
+        labelText = labelText,
+    )
+}
+
+@Composable
+@Deprecated("Use legendItem instead.", ReplaceWith("legendItem(icon, label, labelText)"))
+/**
+ * Defines the appearance of an item of a [VerticalLegend].
+ *
+ * @param icon the [Component] used as the item’s icon.
+ * @param label the [TextComponent] used for the label.
+ * @param labelText the text content of the label.
+ */
+public fun verticalLegendItem(
     icon: Component,
     label: TextComponent,
     labelText: CharSequence,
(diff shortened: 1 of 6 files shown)


vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/HorizontalLegend.kt
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @param items a [Collection] of [Item]s to be displayed by this [HorizontalLegend].
+ * @param iconSizeDp defines the size of all [Item.icon]s.
+ * @param iconPaddingDp defines the padding between each [Item.icon] and its corresponding [Item.label].
+ * @param lineSpacingDp define the vertical spacing between lines.
+ * @param spacingDp defines the horizon spacing between each [Item] in line.
+ * @param padding defines the padding of the content.
+ */
+public open class HorizontalLegend(
+    public var items: Collection<Item>,
```
Gowsky, line 41: The `Item` class is defined in the `VerticalLegend`. Since it is used in both `Legend` implementations, we should move the `Item` into its own top-level class named `LegendItem`.

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/HorizontalLegend.kt
```diff
@@ -0,0 +1,171 @@
...
+    public var iconPaddingDp: Float,
+    public var lineSpacingDp: Float = 0f,
+    public var spacingDp: Float = 0f,
+    override val padding: MutableDimensions = emptyDimensions(),
+) : Legend, Padding {
+    // Cache the height of each line
+    private val heights = mutableListOf<Float>()
+
+    // Cache the items in each line
+    private val lines = mutableListOf<MutableList<Item>>(mutableListOf())
```
Gowsky, line 52: We generally avoid adding comments to code that is self-explanatory. Also, to keep the code style consistent, there should be a new line added after the definition.
```suggestion
    private val lines = mutableListOf<MutableList<Item>>(mutableListOf())
    
```

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/HorizontalLegend.kt
```diff
@@ -0,0 +1,171 @@
...
+    override val bounds: RectF = RectF()
```
Gowsky, line 53:
```suggestion
    override val bounds: RectF = RectF()
    
```

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/HorizontalLegend.kt
```diff
@@ -0,0 +1,171 @@
...
+public open class HorizontalLegend(
+    public var items: Collection<Item>,
+    public var iconSizeDp: Float,
+    public var iconPaddingDp: Float,
+    public var lineSpacingDp: Float = 0f,
+    public var spacingDp: Float = 0f,
+    override val padding: MutableDimensions = emptyDimensions(),
+) : Legend, Padding {
+    // Cache the height of each line
+    private val heights = mutableListOf<Float>()
```
Gowsky, line 49:
```suggestion

    private val heights = mutableListOf<Float>()
```

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/HorizontalLegend.kt
```diff
@@ -0,0 +1,171 @@
...
+            lines.clear()
+            lines.add(mutableListOf())
+            items.forEach {
+                remainWidth -= it.getOriginalWidth(context, availableWidth) - spacingDp.pixels
+                if (remainWidth >= 0 || remainWidth == availableWidth) {
+                    lines[currentLine].add(it)
+                    return@forEach
+                }
+                if (remainWidth < 0) {
+                    // Line break
```
Gowsky, line 135:
```suggestion

```

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/HorizontalLegend.kt
```diff
@@ -0,0 +1,171 @@
...
+    protected open fun Item.getHeight(
+        context: MeasureContext,
+        availableWidth: Float,
+    ): Float = with(context) {
+        label.getHeight(
+            context = context,
+            text = labelText,
+            width = (availableWidth - iconSizeDp.pixels - iconPaddingDp.pixels).toInt(),
+        )
+    }
```
Gowsky, line 152: This is copied from `VerticalLegend`. Once we move the `VerticalLegend.Item` to its own class, as mentioned earlier, this extension function could be moved to the file too. Then we can reuse it and avoid code duplication.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/legend/Legends.kt
```diff
@@ -73,3 +73,53 @@ public fun verticalLegendItem(
...
+    icon: Component,
+    label: TextComponent,
+    labelText: CharSequence,
+): Item = remember(icon, label, labelText) {
+    Item(
+        icon = icon,
+        label = label,
+        labelText = labelText,
+    )
+}
```
Gowsky, line 125: This is a copy of the `verticalLegendItem` function present in the same file. Once we move the `VerticalLegend.Item` to its own top-level class, as mentioned earlier, the `verticalLegendItem` could be renamed to `legendItem`, and this function could be removed.

Gowsky (changes requested):
Hi @lx200916, thank you very much for your contribution! I’ve requested changes that will make the code style more consistent with the existing code, as well as remove code duplication. I’ve also noticed a bug, where a legend item is not moved to the next line.
image

lx200916:
Hi I have moved the `Item` class then rearranged some code.
PS. I failed to reproduce the layout bug but I try to fix some logical errors, could you test it again? 😁

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/Legend.kt
```diff
@@ -35,3 +37,80 @@ public interface Legend : BoundsAware {
...
 }
+
+/**
+ * Defines the appearance of an item of a [Legend].
+ *
+ * @param icon the [Component] used as the item’s icon.
+ * @param label the [TextComponent] used for the label.
+ * @param labelText the text content of the label.
+ */
+public class LegendItem(
```
Gowsky, line 48: The class should be placed in a separate file, as it’s not **directly** related to `Legend` interface, as other implementations may not use the `LegendItem` at all.

vico/core/src/main/java/com/patrykandpatrick/vico/core/legend/Legend.kt
```diff
@@ -35,3 +37,80 @@ public interface Legend : BoundsAware {
...
+    iconPaddingDp: Float,
+    iconSizeDp: Float,
+): Float = with(context) {
+    this@getOriginalWidth.getOriginalLabelWidth(
+        context,
+        availableWidth,
+        iconPaddingDp,
+        iconSizeDp,
+    ) + (iconSizeDp + iconPaddingDp).pixels
+}
```
Gowsky, line 116: All these extension functions could be now member functions.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/legend/Legends.kt
```diff
@@ -62,14 +63,44 @@ public fun verticalLegend(
  * @param labelText the text content of the label.
  */
 @Composable
-public fun verticalLegendItem(
+public fun LegendItem(
```
Gowsky, line 66: As per Jetpack Compose convention, composable function that returns something should start with a lowercase letter.
```suggestion
public fun legendItem(
```

Gowsky (changes requested):

Gowsky:
@lx200916 After your recent changes the bug I’ve experienced is gone. 😉

lx200916:
Hi I have made some changes according to the review, could you check them again?

Gowsky (approved):

lx200916:
😯 Seems like I can not merge it by myself, can you merge it?

Gowsky:
Yeah, sorry for the delay. I and @patrickmichalik were discussing copyright issues and what names you’ve put into your copyright headers. Judging by other libraries (e.g. Accompanist) all source files contain the same copyright headers.
We’d like to know why new files created by yourself have only your name in them (even though they are based on existing code written by us), and the ones that you’ve modified don’t have your name. Was it intentional, or are you open to keeping file headers the same as in already existing files?

lx200916:
😢 Oh I haven't noticed that before🤣 maybe it is generated by some plugins of my IDE, so it's okay to change it to anything else.😆

patrickmichalik (changes requested):
Hello! Thank you for the PR. I’d suggest marking any APIs that we’re shifting away from (such as `verticalLegendItem`) as deprecated, not removing them, to the extent to which this is possible. This way, updating Vico won’t potentially introduce build errors.

lx200916:
> Hello! Thank you for the PR. I’d suggest marking any APIs that we’re shifting away from (such as `verticalLegendItem`) as deprecated, not removing them, to the extent to which this is possible. This way, updating Vico won’t potentially introduce build errors.

Okay but how about the type of param `items` ? Should we keep the `Collection<Item>` or replace it with `Collection<LegendItem>`?

patrickmichalik:
Because `HorizontalLegend` is a new API, it can work with only `LegendItem`. `VerticalLegend`, however, should be compatible with both `LegendItem` and `VerticalLegend.Item`—that is, it should have two constructors (one of them marked as deprecated). I’d suggest introducing an internal function that converts `VerticalLegend.Item` instances to `LegendItem` instances.

lx200916:
> `VerticalLegend`, however, should be compatible with both `LegendItem` and `VerticalLegend.Item`—that is, it should have two constructors (one of them marked as deprecated).

Hey I noticed if we write two constructors whose difference only lies in the type of parameter `items` , it may cause `Platform declaration clash` since they have the same JVM signature due to type erasure(worse: JVMName can not apply to constructors). Static method constructors or some hacky ways may avoid these errors but they seem really weird.

Shall we just add class `VerticalLegend.Item` and then introduce a new compose function `verticalLegend` with `VerticalLegend.Item` which converts `Item` to `LegendItem` internally? Of course they are all marked as deprecated.

Gowsky:
We can solve the JVM signature problem by making the `LegendItem` an `open class`, and making the `VerticalLegend.Item` a subclass of the `LegendItem`. With this change, the `VerticalLegend`’s constructor may accept the `LegendItem`, so it should be binary compatible with older Vico versions.

lx200916:
Hi how about changes like that now?😁

Gowsky:
Thank you very much for your contribution @lx200916!

PR #332, opened 2023-08-03
Added functionality to stick marker to the point

rezafaraji93:
I tried to add the functionality to the marker which the marker can stick to the point and move with it.


patrickmichalik (changes requested):
Hello! Thanks for the PR. Because this is unrelated to `Chart` and its implementations, I think it’d be preferable to keep those unchanged and instead add a parameter to the `MarkerComponent` constructor. Please note that there’s a `markerComponent` composable function, and it would have to be updated too. I’d also suggest using a more conventional name for the parameter—something along the lines of `drawAbovePoint`. Another suggestion would be to position the marker label such that, when `MarkerCorneredShape` is used, the tip of the tick is positioned where the highlighted entry is located, and to have the guideline end at the bottom edge of the label. Finally, please ensure that the code is correctly formatted. You can do so by running the `detekt` Gradle task. Cheers!

patrickmichalik:
Because there hasn’t been any activity here for two months, I’ll be closing the pull request for now. Please feel free to reopen it if you wish to continue working on these changes. Thanks again!

Issue #364, opened 2023-08-18
Texts are too small in Version 1.10.0


p-fischer:
**Describe the bug**
Starting from 1.10.0 all texts in the chart (labels, marker texts, etc...) appear to be much smaller.
In the code, 12sp is defined as default text size. It seems the texts are much smaller than 12sp in version 1.10.0

**Module**

- [x] `compose`
- [ ] `views`
- [ ] `compose-m2`
- [ ] `compose-m3`
- [ ] `core` (select if none of the above apply)

**To reproduce**
Open the vico sample app and compare versions 1.9.2 and 1.10.0

**Expected behavior**
The `textSize` argument is respected (e.g. default value 12sp)

**Screenshots**
Version 1.9.2
![image](https://github.com/patrykandpatrick/vico/assets/8547316/477b561a-3f1f-4853-912a-9ca02fd2d480)

Version 1.10.0
![image](https://github.com/patrykandpatrick/vico/assets/8547316/90584811-6245-4c5f-b125-e09a6c220b99)

**Android versions**
Tested with Android 13

patrickmichalik:
Thanks for the report! [Vico 1.10.1](https://github.com/patrykandpatrick/vico/releases/tag/v1.10.1) resolves this issue.

PR #365, opened 2023-08-18
Fix text sizes

p-fischer:
Text sizes seem to small in version 1.10.0 (see screenshots in #364)

With this fix, text sizes seem to be as big as in version 1.9.2 again.
This was tested on Android 13.

Implementation note:
`convertDimensionToPixels()` is just an alias of `applyDimension()` according to the [official documentation](https://developer.android.com/reference/android/util/TypedValue#convertDimensionToPixels(int,%20float,%20android.util.DisplayMetrics))

### DIFF
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/extension/ContextExtensions.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/extension/ContextExtensions.kt
@@ -17,14 +17,10 @@
 package com.patrykandpatrick.vico.core.extension
 
 import android.content.Context
-import android.os.Build
 import android.util.TypedValue
 
 /**
  * Converts the provided dimension from sp to px.
  */
-public fun Context.spToPx(sp: Float): Float = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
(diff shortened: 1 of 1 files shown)


patrickmichalik:
Hello! Thanks for the PR. We started looking into the bug in question several minutes before you opened the PR, and unfortunately, we noticed the PR only after we’d fixed the problem and started releasing a hotfix ([Vico 1.10.1](https://github.com/patrykandpatrick/vico/releases/tag/v1.10.1)). However, since your solution is more concise, we’d be open to merging your changes. Could you run a rebase to resolve the conflicts that have appeared?

p-fischer:
It's great you reacted so quickly on the issue and released 1.10.1. Thanks!

I rebased this PR, which now is not an urgent bugfix anymore but should be a pure refactoring.

patrickmichalik (changes requested):
Thanks! The Detekt workflow is failing because `import android.os.Build` is redundant. Could you remove that?

p-fischer:
> Thanks! The Detekt workflow is failing because `import android.os.Build` is redundant. Could you remove that?

Sure! Sorry about that.

patrickmichalik (approved):
No worries! Everything looks good now. Thanks!

Gowsky (approved):

Issue #387, opened 2023-09-04
Label count off when chart has negative values


p-fischer:
**Describe the bug**
I am creating an instance of `DefaultVerticalAxisItemPlacer` with `maxItemCount` = 8 and `getMixedLabelValues()` returns 9 values. Here is a screenshot of the function with a breakpoint in the end:
![image](https://github.com/patrykandpatrick/vico/assets/8547316/191a32b6-bb3f-41e9-bb22-caa9958a3355)

When I comment out the following block, the code works as expected:
```
if (maxTopItemCount % 1f != 0f) {
    val isTopNotDenser = topItemCount / topHeight <= bottomItemCount / bottomHeight
    val isTopFillable = topItemCountByHeight - topItemCount >= 1
    val isBottomFillable = bottomItemCountByHeight - bottomItemCount >= 1
    if (isTopFillable && (isTopNotDenser || !isBottomFillable)) {
        topItemCount++
    } else if (isBottomFillable) {
        bottomItemCount++
    }
}
```

The code block is entered here because `maxItemCount` is `5.0000005`. 

My guess is that the above code wants to fix a case where the result would be one label short. However, this is not the case here: `topItemCount` = `5`, `bottomItemCount`= `2` plus the zero label equals `maxItemCount`.

A fix could be to change the `if` check from `maxTopItemCount % 1f != 0f` to `topItemCount + bottomItemCount + 1 < maxItemCount`

**Module**

- [x] `compose`
- [ ] `views`
- [ ] `compose-m2`
- [ ] `compose-m3`
- [ ] `core` (select if none of the above apply)

**To reproduce**
You need a chart with negative values and I used rounded max and min label values according to the description here: https://github.com/patrykandpatrick/vico/issues/224#issuecomment-1613568847.
If I change the top padding of the chart the bug dissapears so it's not always reproducable but depends on round issues due to float imprecision.

**Screenshots**
Result with the original code:
![image](https://github.com/patrykandpatrick/vico/assets/8547316/7bdea099-fa7c-474e-af87-28edf252f043)

Result with the suggested code change:
![image](https://github.com/patrykandpatrick/vico/assets/8547316/c432ab6b-1c00-41b8-bfc8-0b91e58a363d)

**Android versions**
Tested with Android 13

I am going to create a pull request.

patrickmichalik:
Hello! Thanks for the report. I’ll be taking a look at the PR in a second.

patrickmichalik:
[...]

PR #388, opened 2023-09-04
Charts with negative values can show one label more than maxItemCount permits

p-fischer:
addresse #387


patrickmichalik (changes requested):
Thanks for the PR! The fix looks good. However, as reported by Detekt, the position of the comment isn’t optimal. Could you please move it?

patrickmichalik (approved):
Thanks!

Gowsky (approved):

Issue #369, opened 2023-08-22
Using a ChartEntryModelProducer breaks Compose preview


Rosomack:
**Describe the bug**

When using a `ChartEntryModelProducer` charts do not render at all in the Compose preview (they do take up space though). The chart renders fine on a device.

It also works fine when providing an entry model instead.

**Module**

- [x] `compose`
- [ ] `views`
- [ ] `compose-m2`
- [ ] `compose-m3`
- [ ] `core` (select if none of the above apply)

**To reproduce**
Steps to reproduce the behavior.

1. Create a `ChartEntryModelProducer` instantiating it with the preview data
2. Create a chart using this `ChartEntryModelProducer`
3. Create a compose preview for the above
4. Chart does not render in the preview

**Expected behavior**

Chart preview should render in the preview.

**Screenshots**

**Android versions**
Versions of Android where the bug occurs.

Android Studio Giraffe | 2022.3.1

**Additional context**

Leedwon:
Hi, I think you can create preview with `Chart` and `ChartEntryModelProducer`. I'm using Vico 1.9.2 and the following code works for me:
```Kotlin
@Composable
private fun Chart(
    chartEntryModelProducer: ChartEntryModelProducer
) {
    ProvideChartStyle(rememberChartStyle()) {
        Chart(
            chart = lineChart(),
            chartModelProducer = chartEntryModelProducer,
            startAxis = rememberStartAxis(),
            bottomAxis = rememberBottomAxis(guideline = null),
            runInitialAnimation = false
        )
    }
}

@Composable
@Preview
private fun Preview() {
    Surface {
        Chart(
            ChartEntryModelProducer(
                listOf(
                    FloatEntry(0.0f, 0.0f),
                    FloatEntry(1.0f, 2.0f),
                    FloatEntry(2.0f, 3.0f),
                    FloatEntry(3.0f, 4.0f),
                    FloatEntry(4.0f, 5.0f),
                )
            )
        )
    }
}
```

I'm able to see the following preview:
![preview_vico](https://github.com/patrykandpatrick/vico/assets/29312085/bd2014ca-b145-445c-9d3f-35d413b24098)

Rosomack:
@Leedwon Thanks, I couldn't find this as I was on an earlier version of the library. Appreciate the help!

In case it helps anyone you can use 

`runInitialAnimation = !LocalInspectionMode.current`

[...]

PR #389, opened 2023-09-04
Disable animations in previews

Leedwon:
This PR aims to resolve [#369](https://github.com/patrykandpatrick/vico/issues/369) by disabling animations in previws using `LocalInspectionMode.current`. One thing to note here is that as mentioned in #369 it will also disable animations in interactive previews. I was looking for a solution to detect whether preview is interactive or not, but with no luck. My rationale behind this change is that non interactive preview is shown by default and currently preview seems to be broken when using `ChartEntryModelProducer`. This PR introduces a trade-off ensuring that preview is working in non-interactive mode, but breaking interactive preview by default, since there will be no animation. I think you have to decide which behaviour is more preferable for you. If you decide to merge it, I think it would be worth to add information about interactive previews somewhere in the docs/README/kdoc, if you point me to the right spot I can do it as well.


Gowsky (approved):

patrickmichalik (approved):
Hello! Thanks for the PR. This looks good. Don’t worry about the documentation—we’ll take care of it.

PR #650, opened 2024-04-08
Expose TextComponent functions which accept androidx.compose.ui.text.TextStyle

kevincianfarini:
The functions `rememberTextComponent` and `rememberAxisLabelComponent`
now both accept a `TextStyle` that dictates the text's color, size,
and font. Previously, these functions accepted three parameters for
these values. This was awkward for Compose callers, particularly for
the `android.graphics.Typeface` value.

### DIFF
--- a/sample/src/main/java/com/patrykandpatrick/vico/sample/previews/AxisLinePreviews.kt
+++ b/sample/src/main/java/com/patrykandpatrick/vico/sample/previews/AxisLinePreviews.kt
@@ -18,6 +18,7 @@ package com.patrykandpatrick.vico.sample.previews
 
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.patrykandpatrick.vico.compose.cartesian.CartesianChartHost
@@ -46,6 +47,7 @@ private val model = CartesianChartModel(ColumnCartesianLayerModel.build { series
 fun HorizontalAxisTextInside() {
     val label =
         rememberAxisLabelComponent(
+            style = TextStyle.Default,
             background =
                 rememberShapeComponent(
                     shape =
@@ -101,6 +103,7 @@ fun HorizontalAxisTextInside() {
 fun HorizontalAxisTextInsideAndBottomAxis() {
     val label =
         rememberAxisLabelComponent(
+            style = TextStyle.Default,
             background =
                 rememberShapeComponent(
                     shape = Shape.Pill,
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/common/component/Components.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/common/component/Components.kt
@@ -25,13 +25,16 @@ import androidx.compose.runtime.Stable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.isSpecified
 import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.TextUnit
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.patrykandpatrick.vico.compose.common.pixelSize
 import com.patrykandpatrick.vico.compose.common.shader.toDynamicShader
+import com.patrykandpatrick.vico.compose.common.text.toGraphicsTypeFace
 import com.patrykandpatrick.vico.core.common.Defaults
 import com.patrykandpatrick.vico.core.common.Dimensions
 import com.patrykandpatrick.vico.core.common.LayeredComponent
@@ -154,6 +157,9 @@ public fun rememberLayeredComponent(
  * @param textAlignment the text alignment.
  * @param minWidth defines the minimum width.
  */
+@Deprecated(
+    message = "Use the Composable function that accepts a Compose UI TextStyle instead of an android graphics Typeface.",
+)
 @Composable
 public fun rememberTextComponent(
     color: Color = Color.Black,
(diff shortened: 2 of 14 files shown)


vico/compose/src/main/java/com/patrykandpatrick/vico/compose/common/component/Components.kt
```diff
@@ -182,6 +188,47 @@ public fun rememberTextComponent(
...
+    lineCount: Int = Defaults.LABEL_LINE_COUNT,
+    padding: MutableDimensions = MutableDimensions.empty(),
+    margins: MutableDimensions = MutableDimensions.empty(),
+    textAlignment: Layout.Alignment = Layout.Alignment.ALIGN_NORMAL,
+    minWidth: TextComponent.MinWidth = TextComponent.MinWidth.fixed(),
+): TextComponent {
+    val typeface = textStyle.toGraphicsTypeFace()
+    return remember(textStyle, typeface, background, ellipsize, lineCount, padding, margins, typeface, textAlignment, minWidth) {
+        TextComponent.build {
+            this.color = if (textStyle.color.isSpecified) textStyle.color.toArgb() else Color.Black.toArgb()
```
kevincianfarini, line 218: I was unsure if we should expose a secondary function parameter which allows configuring the Color of this text. If it's desirable, it would probably look like this: 

```kt
@Composable
fun rememberTextComponent(
  textStyle: TextStyle = TextStyle.Default, 
  color: Color = if (textStyle.color.isSpecified) textStyle.color else Color.Black, 
  // Other params...
)
```

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/common/component/Components.kt
```diff
@@ -154,6 +157,9 @@ public fun rememberLayeredComponent(
  * @param textAlignment the text alignment.
  * @param minWidth defines the minimum width.
  */
+@Deprecated(
+    message = "Use the Composable function that accepts a Compose UI TextStyle instead of an android graphics Typeface.",
+)
```
kevincianfarini, line 162: I notice we don't use Kotlin's `ReplaceWith` in other areas of the codebase. Should I omit them here?

patrickmichalik (changes requested):
Hello, and thanks for the PR!

1. It looks like the Ktlint workflow is failing. Could you reformat the code?
2. I think it would be preferable to have several parameters of Compose types rather than a `TextStyle` parameter. `TextStyle` has over a dozen properties, but only four are read here. It may be unintuitive for some properties to be silently ignored. I apologize for not mentioning this [here](https://github.com/patrykandpatrick/vico/discussions/641#discussioncomment-9037553)—it caught my attention just now.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/common/component/Components.kt
patrickmichalik: Please see the second point [here](https://github.com/patrykandpatrick/vico/pull/650#pullrequestreview-1988288337). This change would involve adding a `color` parameter.

kevincianfarini: Okay! Shall I expose a text size parameter here, too?

kevincianfarini: I went ahead and exposed a size parameter. Will push momentarily.

kevincianfarini:
I've pushed some edits. They: 

* Add `color` and `size` to both of the composable functions. The defaults for these values correspond to the value in the `TextStyle`. 
* Reformatted the code to adhere to lint. 

CI won't run until you approve, so consider this a ping 😄

sample/src/main/java/com/patrykandpatrick/vico/sample/previews/AxisLinePreviews.kt
```diff
@@ -46,6 +47,7 @@ private val model = CartesianChartModel(ColumnCartesianLayerModel.build { series
 fun HorizontalAxisTextInside() {
     val label =
         rememberAxisLabelComponent(
+            style = TextStyle.Default,
```
kevincianfarini, line 50: The two overloads of `rememberTextComponent` and `rememberAxisLabelComponent` have conflicting overloads when not specifying certain parameters. Rather than abruptly removing the deprecated declarations, I opted to manually specify the default value in all of the call sites in the sample project which experienced this ambiguity. 

I don't love this. Should we consider abrupt breakage for this since Vico 2 is still in alpha?

Gowsky:
Hello @kevincianfarini, thanks for your contribution! Regarding @patrickmichalik’s comment, he meant that we shouldn't use `TextStyle` as a parameter. That’s because we won’t use most of its properties. Instead, we should directly accept `FontFamily`, `FontWeight`, `FontStyle`, and `FontSynthesis`.

kevincianfarini:
Ah I see, I can rework that.

kevincianfarini:
I'm going to have to follow up on this next week since I'm going to be out of the office. Sorry for the delay!

patrickmichalik:
No worries at all, @kevincianfarini! There’s no rush here. Thanks for letting us know.

sample/src/main/java/com/patrykandpatrick/vico/sample/previews/AxisLinePreviews.kt
patrickmichalik: Apologies for the late reply. If you’re still interested in finalizing this pull request, I’d suggest updating the existing overloads such that the `typeface` parameters don’t have default values. The conflicts should then disappear.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/common/component/Components.kt
patrickmichalik: Sounds good. Thanks!

Gowsky:
[Ping](https://github.com/patrykandpatrick/vico/discussions/621#discussioncomment-9135988). 😄

kevincianfarini:
Hey, sorry, things have gotten extremely busy with work. I will try to get this over the finish line in the next couple of weeks, though if you'd prefer please feel free to take this over. I won't be offended.

Gowsky:
No problem. Thanks for the update!

Gowsky:
Closing due to inactivity.

PR #677, opened 2024-04-27
Dashed Path support for Line Layer

anshu7vyas:
# Description
This pull request enables users to add a `android.graphics.PathEffect` of their choice to the `LineCartesianLayer`.

# Why?
On occasion, the design team may require a dashed pattern for the lines on the Line Layer. There's also an ongoing discussion about adding support for a Dashed Line 
- #469 

With this update, users can add any `PathEffect` they prefer, allowing them more flexibility than just a dashed effect.

# How?
First, I added a nullable `pathEffect` property to the `LineSpec` in `/core/cartesian/layer/LineCartesianLayer.kt` class. Utilizing the library's extensive APIs, I assigned this `pathEffect` to `linePaint` within the `drawLine()` function.

Next, I updated the `rememberLineSpec()` function in `/compose/cartesian/layer/LineCartesianLayer.kt` to include the `pathEffect` property.

And finally, I updated the sample application by using the `pathEffect` property on `Chart9`. This shows how simple it is to add a new line effect.

# What does that look like?

| Default Behavior | with `DashPathEffect` |
| ---| ---|
|  |  |

| Default Behavior | with `DashPathEffect` |
| ---| ---|
| <video src="https://github.com/patrykandpatrick/vico/assets/9023987/72a197f8-8c40-462c-a91f-ec347617f077"/> | <video src="https://github.com/patrykandpatrick/vico/assets/9023987/cd0eb0d8-8fef-4cba-89fc-6fb3ac4085a1"/> |

### DIFF
--- a/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/charts/Chart9.kt
+++ b/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/charts/Chart9.kt
@@ -16,6 +16,7 @@
 
 package com.patrykandpatrick.vico.sample.showcase.charts
 
+import android.graphics.DashPathEffect
 import android.graphics.PorterDuff
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
@@ -100,6 +101,8 @@ private fun ComposeChart9(
 ) {
     val colors = chartColors
     val marker = rememberMarker()
+    // Set dashed line style
+    val dashPathEffect = DashPathEffect(floatArrayOf(PATH_EFFECT_DASH_LENGTH, PATH_EFFECT_DASH_GAP), 0f)
     CartesianChartHost(
         chart =
             rememberCartesianChart(
@@ -146,6 +149,7 @@ private fun ComposeChart9(
                                             PorterDuff.Mode.DST_IN,
                                         ),
                                     ),
+                                pathEffect = dashPathEffect,
                             ),
                         ),
                 ),
@@ -207,6 +211,8 @@ private fun ViewChart9(
 ) {
     val marker = rememberMarker()
     val colors = chartColors
+    // Set dashed line style
+    val dashPathEffect = DashPathEffect(floatArrayOf(PATH_EFFECT_DASH_LENGTH, PATH_EFFECT_DASH_GAP), 0f)
     AndroidViewBinding(Chart9Binding::inflate, modifier) {
         with(chartView) {
             runInitialAnimation = false
@@ -252,6 +258,7 @@ private fun ViewChart9(
                                         PorterDuff.Mode.DST_IN,
                                     ),
                                 ),
+                            pathEffect = dashPathEffect,
                         ),
                     )
             }
@@ -269,3 +276,6 @@ private val chartColors
         )
 
 private val x = (1..100).toList()
+
+private const val PATH_EFFECT_DASH_LENGTH = 20f // Length of each dash
+private const val PATH_EFFECT_DASH_GAP = 10f // Length of the gap between dashes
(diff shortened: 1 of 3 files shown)


patrickmichalik (changes requested):
Hello, and thanks for the PR! Combining `CornerPathEffect` and a `LineCartesianLayer.LineSpec.PointConnector` that uses `lineTo` produces the following. Please note the mismatch between the line and the background.



Updating the code to apply `LineCartesianLayer.LineSpec#pathEffect` to `LineCartesianLayer.LineSpec#lineBackgroundPaint`, we get the following. Please take a look at the start and the end of the line.



Achieving full `PathEffect` compatibility may be a bit challenging. If you have an idea and would like to implement it, please feel free to do so. Otherwise, we’d suggest adding dedicated dash properties and using `DashPathEffect` internally.

`LineCartesianLayer` is optimized to draw only the visible part of each line. With dashed lines, however, this causes the dashes to shift:

https://github.com/patrykandpatrick/vico/assets/120058021/4828bdb0-8061-48e9-8f1e-5a2a8525af24

We’d thus suggest modifying the logic such that dashed lines are drawn from the very start and cut off only past the viewport’s end edge. Another option is to cut off dashed lines on both sides (as with continuous lines) and shift the dashes as required, most likely using `PathMeasure` to run the calculations. However, this would be significantly more complex, and we’re not entirely sure if it would perform discernibly better. The former solution is perfectly sufficient for now. The logic that determines for what points lines should be drawn is in `CartesianDrawContext#forEachPointInBounds` (in `LineCartesianLayer`).

With the ninth sample chart being rather complex, we’d suggest leaving it as is and showcasing this feature in, for example, the third sample chart.

anshu7vyas:
Hi @patrickmichalik, thanks for reviewing the PR and offering insightful feedback. Here's my interpretation of your suggestions:

> we’d suggest adding dedicated dash properties and using DashPathEffect internally.
> 

To address this, I'll adjust the exposed `pathEffect` property to be specifically `DashPathEffect` with default dash properties within the library. This will allow users to modify the `dashPathLength` and `dashPathGap` for the effect. Alternatively, would you recommend adding a simple boolean `isDashedPathEffectEnabled` instead?

> We’d thus suggest modifying the logic such that dashed lines are drawn from the very start and cut off only past the viewport’s end edge.

Could you provide a bit more clarification on this point?

> With the ninth sample chart being rather complex, we’d suggest leaving it as is and showcasing this feature in, for example, the third sample chart.

I'll update the sample project to demonstrate `DashPathEffect` in `Chart3.kt`.

patrickmichalik:
I’d suggest creating a sealed interface and updating `LineCartesianLayer.LineSpec` to accept an implementation thereof, @anshu7vyas. The interface would have two implementations, one for continuous lines and one for dashed lines. The latter would have parameters for the dash length and gap length. This would enable us to accept measurements expressed in density-independent pixels—`DashPathEffect` takes pixel measurements. As with `HorizontalLayout`, for example, factory functions would be added to the `compose` module. The function that instantiates the dashed-line implementation of the sealed interface would have `Dp` parameters.

Regarding the issue where the dashes shift, the following recording should make things clearer. I’ve disabled clipping here, so we can see what happens outside of the `LineCartesianLayer` bounds.

https://github.com/patrykandpatrick/vico/assets/120058021/043ad8d2-c770-4619-949f-56953d77524f

For the shifting issue not to occur, dashed lines can be drawn as follows:

https://github.com/patrykandpatrick/vico/assets/120058021/41b0419a-ff09-4580-b448-80472a87ea78

The condition [here](https://github.com/patrykandpatrick/vico/blob/6516a10de23338cf56168b4cb494c1335e1249fa/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt#L517-L518) can be updated to always be false for dashed lines.

patrickmichalik:
You’re welcome, @anshu7vyas! Sounds good. Cheers!

Gowsky:
I'm closing this due to 2 months of inactivity. Thanks for your contribution @anshu7vyas!

PR #679, opened 2024-04-29
AxisVertical: Inner Top label on position 0

Antiglobalist:
Display the first label vertically within if the VerticalLabelPosition is set to Top.

Example of the required Chart
image

After applying the styles and with this change
image


Gowsky:
Hello @Antiglobalist, thank you for your contribution. We've identified some problems with the approach you've chosen:

1. Incompatibility with top axis.

    image

1. There's also the opposite problem with `VerticalLabelPosition.Bottom`. 15 should be displayed here, so the fix works both ways.

    image

This would have to be rewritten, so it's more universal. Would you be interested?

PR #870, opened 2024-09-05
Dashed Path support for LineCartesianLayer

anshu7vyas:
# Description
This pull request enables users to add a `android.graphics.DashPathEffect` to the `LineCartesianLayer`. It adds a `LinePattern` sealed interface for `LineCartesianLayer` to add dashed line pattern support along with continuous line pattern.

# Why?
Occasionally, the design team may require a dashed pattern for the lines on the Line Layer. There's also an ongoing discussion about adding support for a Dashed Line 
- Previous attempt: #677 , my apologies for not being able to address the comments on time.
- Open discussion: #469 

With this update, users can choose a `LinePattern` between `Continuous` (default) and `Dashed`

# How?
First, I added a sealed interface `LinePattern` containing two implementation `Continuous` and `Dashed` addressing this [comment](https://github.com/patrykandpatrick/vico/pull/677#issuecomment-2105937285) from the previous PR. Then I added it as a property for `Line` in `/core/cartesian/layer/LineCartesianLayer.kt` class. Utilizing the library's extensive APIs, I assigned these patterns to `pathEffect` using `linePaint` within the `draw()` function.

Next, I updated the `rememberLine()` function in `/compose/cartesian/layer/LineCartesianLayer.kt` to include the `pattern` property. I also added compose factory methods for `continuous()` and `dashed()` so that we can use `Dp` in Compose, addressing this [comment](https://github.com/patrykandpatrick/vico/pull/677#issuecomment-2105937285) 

And finally, I updated the sample application by using the `pattern` property on `Chart3` (as suggested [here](https://github.com/patrykandpatrick/vico/pull/677#pullrequestreview-2035396974)). This shows how simple it is to add a new line pattern.

# What does that look like?

| Default Behavior | with `Dashed` line pattern |
| ---| ---|
|  |  |

| Default Behavior | with `Dashed` line pattern |
| ---| ---|
| <video src="https://github.com/user-attachments/assets/bdcaaba8-2591-470e-9c8f-f459d03e8b2e"/> | <video src="https://github.com/user-attachments/assets/010a339e-f9ab-43eb-8af7-81761b35ad81"/> |

<details>

<summary>Known issues ✅ </summary>

# Known issues:
I have encountered a couple of issues that I haven't been able to fully resolve:

- ✅  **Shifting dashes on scroll**: As pointed out by @patrickmichalik in this [comment](https://github.com/patrykandpatrick/vico/pull/677#issuecomment-2105937285), `LineCartesianLayer` optimizes performance by rendering only the visible portion of each line. While this works well for continuous lines, it causes the dashes to shift when scrolling, as the dash pattern resets based on the visible section. Given my limited understanding of the APIs, ~I haven't found a way to fix this yet.~ Fixed in [1817969](https://github.com/patrykandpatrick/vico/pull/870/commits/181796993ef1a09e6832ebc9857f143a4bc30ada)

**This issue can be ignored as the fix is already being made by the authors - see [comment](https://github.com/patrykandpatrick/vico/pull/870#issuecomment-2351111820)**
>~- ✅  **Markers Falling on Dash Gaps**: While updating `Chart3`, I noticed that the `Marker` can appear empty when it falls on a dash gap. This occurs because the marker's position lands where no path is drawn. A possible fix could involve using `PathDashPathEffect` to draw a transparent underlying path with the dashed path overlaid on top, ensuring the marker always has a valid Entry to read the value from. However, I'm unsure if this approach is optimal or merely a workaround.
~

I’d appreciate any guidance or feedback on how to best address these issues, as they are crucial to ensuring smooth functionality. I understand this PR is incomplete without addressing these, I intend to get the ball rolling to add this functionality to this already amazing library. 🙌  If you plan to address this your way, please feel free to close this PR in favor of your implementation.

</details>

Please feel free to make any changes as you see fit!

Thank you!

### DIFF
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/layer/LineCartesianLayer.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/layer/LineCartesianLayer.kt
@@ -95,9 +95,8 @@ public fun LineCartesianLayer.Companion.rememberLine(
     vicoTheme.lineCartesianLayerColors.first().let { color ->
       remember(color) { LineCartesianLayer.LineFill.single(fill(color)) }
     },
-  thickness: Dp = Defaults.LINE_SPEC_THICKNESS_DP.dp,
+  stroke: LineCartesianLayer.LineStroke = LineCartesianLayer.LineStroke.continuous(),
   areaFill: LineCartesianLayer.AreaFill? = remember(fill) { fill.getDefaultAreaFill() },
-  cap: StrokeCap = StrokeCap.Round,
   pointProvider: LineCartesianLayer.PointProvider? = null,
   pointConnector: LineCartesianLayer.PointConnector = remember {
     LineCartesianLayer.PointConnector.cubic()
@@ -109,9 +108,8 @@ public fun LineCartesianLayer.Companion.rememberLine(
 ): LineCartesianLayer.Line =
   remember(
     fill,
-    thickness,
+    stroke,
     areaFill,
-    cap,
     pointProvider,
     pointConnector,
     dataLabel,
@@ -121,9 +119,8 @@ public fun LineCartesianLayer.Companion.rememberLine(
   ) {
     LineCartesianLayer.Line(
       fill,
-      thickness.value,
+      stroke,
       areaFill,
-      cap.paintCap,
       pointProvider,
       pointConnector,
       dataLabel,
@@ -150,3 +147,24 @@ private val StrokeCap.paintCap: Paint.Cap
           "Not `StrokeCap.Butt`, `StrokeCap.Round`, or `StrokeCap.Square`."
         )
     }
+
+/** Creates a [LineCartesianLayer.LineStroke.Continuous] instance. */
+public fun LineCartesianLayer.LineStroke.Companion.continuous(
+  thickness: Dp = Defaults.LINE_SPEC_THICKNESS_DP.dp,
+  cap: StrokeCap = StrokeCap.Round,
+): LineCartesianLayer.LineStroke.Continuous =
+  LineCartesianLayer.LineStroke.Continuous(thickness.value, cap.paintCap)
+
+/** Creates a [LineCartesianLayer.LineStroke.Dashed] instance. */
+public fun LineCartesianLayer.LineStroke.Companion.dashed(
+  thickness: Dp = Defaults.LINE_SPEC_THICKNESS_DP.dp,
+  cap: StrokeCap = StrokeCap.Round,
+  dashLength: Dp = Defaults.LINE_PATTERN_DASHED_LENGTH.dp,
+  gapLength: Dp = Defaults.LINE_PATTERN_DASHED_GAP.dp,
+): LineCartesianLayer.LineStroke.Dashed =
+  LineCartesianLayer.LineStroke.Dashed(
+    thickness.value,
+    cap.paintCap,
+    dashLength.value,
+    gapLength.value,
+  )
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt
@@ -18,9 +18,9 @@ package com.patrykandpatrick.vico.core.cartesian.layer
 
 import android.graphics.Bitmap
 import android.graphics.Canvas
+import android.graphics.DashPathEffect
 import android.graphics.Paint
 import android.graphics.Path
-import android.graphics.PorterDuff
 import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.Stable
 import com.patrykandpatrick.vico.core.cartesian.CartesianDrawingContext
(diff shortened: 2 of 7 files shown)


Gowsky:
Hello, @anshu7vyas. Thank you for your contribution.

>Shifting dashes on scroll: As pointed out by @patrickmichalik in this https://github.com/patrykandpatrick/vico/pull/677#issuecomment-2105937285, LineCartesianLayer optimizes performance by rendering only the visible portion of each line. While this works well for continuous lines, it causes the dashes to shift when scrolling, as the dash pattern resets based on the visible section. Given my limited understanding of the APIs, I haven't found a way to fix this yet.

@patrickmichalik shared how to prevent this from happening in the comment you have linked. It's at the very bottom. Do you have any specific questions about this solution?

>Markers Falling on Dash Gaps: While updating Chart3, I noticed that the Marker can appear empty when it falls on a dash gap. This occurs because the marker's position lands where no path is drawn. A possible fix could involve using PathDashPathEffect to draw a transparent underlying path with the dashed path overlaid on top, ensuring the marker always has a valid Entry to read the value from. However, I'm unsure if this approach is optimal or merely a workaround.

A change we plan on making will prevent this from happening. So it can be ignored for now.

anshu7vyas:
Thanks @Gowsky, appreciate the feedback.
Re: shifting dashes on scroll: 
> @patrickmichalik shared how to prevent this from happening in the comment you have linked. It's at the very bottom. Do you have any specific questions about this solution?

I understood where to make the change but I am unsure how to make this condition `return false` for `dashed` line pattern as we don't expose the `linePattern` property outside of `Line` class. I thought about extracting `linePaint` property outside of `Line` class but that would mean we would have to remove `cap: Paint.Cap` property and default it to `Paint.Cap.ROUND` for `linePaint` instantiation alongside `linePath`. If that is feasible, I should then be able to check for this [condition](https://github.com/patrykandpatrick/vico/blob/6516a10de23338cf56168b4cb494c1335e1249fa/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt#L517-L518) if the `pathEffect` for this `linePaint` is `null`, we keep the optimization, otherwise return false. 

Please let me know if I am missing something, thank you!

Gowsky:
>I understood where to make the change but I am unsure how to make this condition return false for dashed line pattern as we don't expose the linePattern property outside of Line class.

I'd suggest adding a `requiresFullLineLength: Boolean` property to the `Line` class that would return `true` when `Line.pattern` is set to `LinePattern.Dashed`. [`LineCartesianLayer.forEachPointInBounds`](https://github.com/patrykandpatrick/vico/blob/6516a10de23338cf56168b4cb494c1335e1249fa/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt#L479) would need a new parameter `drawFullLineLength: Boolean`. When the parameter is `true`, the `forEachPointInBounds` function would skip the optimization and draw the line from the start. `drawFullLineLength` would need to be set to `Line.requiresFullLineLength` only when it's used to draw the line. All other usages must use `false`.

anshu7vyas:
> I'd suggest adding a `requiresFullLineLength: Boolean` property to the `Line` class that would return `true` when `Line.pattern` is set to `LinePattern.Dashed`. [`LineCartesianLayer.forEachPointInBounds`](https://github.com/patrykandpatrick/vico/blob/6516a10de23338cf56168b4cb494c1335e1249fa/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt#L479) would need a new parameter `drawFullLineLength: Boolean`. When the parameter is `true`, the `forEachPointInBounds` function would skip the optimization and draw the line from the start. `drawFullLineLength` would need to be set to `Line.requiresFullLineLength` only when it's used to draw the line. All other usages must use `false`.

Thank you so much for your input and solution, @Gowsky 🙌 . I have made this change in [1817969](https://github.com/patrykandpatrick/vico/pull/870/commits/181796993ef1a09e6832ebc9857f143a4bc30ada), please take a look. 
Here's a before-after video for the change:

| Before | After |
| ---| ---|
| <video src="https://github.com/user-attachments/assets/ebdde07c-b186-4a86-afd1-99f4c826c08f"/> | <video src="https://github.com/user-attachments/assets/0850395c-a6b7-4702-b10e-d15d3c5dbf16"/> |

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt
```diff
@@ -129,6 +135,13 @@ public open class LineCartesianLayer(
         val thickness = thicknessDp.pixels
         linePaint.strokeWidth = thickness
         val halfThickness = thickness.half
+        // Apply the PathEffect based on the line pattern
```
Gowsky, line 138: Please remove this comment, as the code is self-explanatory.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/layer/LineCartesianLayer.kt
```diff
@@ -73,6 +73,9 @@ public fun LineCartesianLayer.Companion.rememberLine(
     vicoTheme.lineCartesianLayerColors.first().let { color ->
       remember(color) { LineCartesianLayer.LineFill.single(fill(color)) }
     },
+  pattern: LineCartesianLayer.LinePattern = remember {
+    LineCartesianLayer.LinePattern.continuous()
+  },
```
Gowsky, line 78: There's no point in remembering an `object`.

Gowsky (changes requested):
Please reformat the code with ktfmt using `google-style`. Also, we need to have XML attributes for using the dashed path effect.

Gowsky:
Closing due to inactivity.

Gowsky (approved):

Gowsky:
Since this was almost merge-ready, we decided to reopen and update it. We will finalize it.

Gowsky:
Thanks for your contribution, @anshu7vyas!

Issue #900, opened 2024-09-30
2.0.0-beta.1 Introduces Failure to Recompose Line


Tyler-Lopez:
### How to reproduce

Hi there!

I believe that `2.0.0-beta.1` introduces a new bug. I have included reproducible examples and videos. 

When a property of a `LineCartesianLayer` is changed, the style _should_ recompose.

In `2.0.0-alpha.28`, if I have a state parameter that is used for the thickness of the line, when it changes, I successfully see the line update.

This gist demonstrates a successful case in `2.0.0-alpha.28`.

https://gist.github.com/Tyler-Lopez/39f0768ce9dd2c0aec8c281aa702033b

If I take the same code and update my version to `2.0.0-beta1`, it no longer successfully updates. 

This gist demonstrates a failure case in `2.0.0-beta1`.

https://gist.github.com/Tyler-Lopez/7ca8630b7bb8e7bee26d73dfb964cbb9

I was drawn to pursue this as a possible reason for the problem as part of the change between the versions is how `rememberLine` is invoked. I have tried not using that function in the beta and was also unsuccessful.

### Observed behavior

https://github.com/user-attachments/assets/aeef03e2-86d5-4496-89a1-72fd1a166f3a

### Expected behavior

https://github.com/user-attachments/assets/1c114584-cf5d-44eb-a060-b9869fc7c4d6

### Vico version(s)

2.0.0-beta.1

### Android version(s)

33

### Additional information

I think it would be great if one of the "example charts" on the showcase app featured style state-mutation like this. Maybe this is something I could contribute towards? 

In my app, I am working on developing such a "showcase page" to make sure the chart always works as expected.

Gowsky:
Hello, @Tyler-Lopez. This is caused by strong skipping mode being enabled by default in Kotlin 2.0.20. We are looking into it.

PR #913, opened 2024-10-09
Feature/marker on tap event

cristianboicu01:
### Description

This pr aims to add support for onTap events in order to track the last tap point on the chart. Also, it makes possible for a marker to be displayed only when a tap event occurs which was previously not achievable.

### Why?

This feature can be useful when a user needs to display a marker only on tap event and wants to avoid marker popping up on every touch event. For instance, such unwanted behaviour may occur while scrolling. 

For context: [#769 ](https://github.com/patrykandpatrick/vico/discussions/769#discussioncomment-10834453)

### How?

1. Adds a `displayOnTap` value inside `CartesianMarker` which by default is `false`. This can be changed only by creating a custom implementation of a marker and setting the value to `true`. Tap detection will work only if the `displayOnTap` is `true`.

2. Gesture detection

- For **Compose**, detect tap gesture by adding an another `pointerInput` modifier inside `TouchEvent`. Tap event is sent further through a dedicated callback.
- For **Android View,** detect tap gesture inside `MotionEventHandler` by listening for `MotionEvent.ACTION_UP` and checking the drag distance. Like before, tap event is sent further through a dedicated callback.
- Calls `chart.draw()` replacing `markerTouchPoint` with `markerTapPoint`.
- **Important thing to note is that tap event and touch event are mutual exclusive.** 

3. Drawing marker
- Adds `displayMarkerOnTap` parameter to `chart.draw()` and also to `drawMarker()`. It is used to figure out how and when the marker should be drawn and when to notify user about it. 
- `lastTapPosition` helps to avoid redrawing the marker on new chart item when user scrolls.

4. CartesianMarkerVisibilityListener
- Adds a `onTap(marker: CartesianMarker, targets: List<CartesianMarker.Target>)` callback to notify user that marker was drawn on a new position.

5. Adds `Chart11` to showcase this behaviour. 

**Note:**
In Android View version of `Chart11`, persistent markers do not update properly. Might be related to [#900](https://github.com/patrykandpatrick/vico/issues/900)

It may not be the most elegant solution but it does the job.

### DIFF
--- a/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/Charts.kt
+++ b/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/Charts.kt
@@ -20,6 +20,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import com.patrykandpatrick.vico.sample.showcase.charts.Chart1
 import com.patrykandpatrick.vico.sample.showcase.charts.Chart10
+import com.patrykandpatrick.vico.sample.showcase.charts.Chart11
 import com.patrykandpatrick.vico.sample.showcase.charts.Chart2
 import com.patrykandpatrick.vico.sample.showcase.charts.Chart3
 import com.patrykandpatrick.vico.sample.showcase.charts.Chart4
@@ -41,4 +42,5 @@ internal val charts =
     { uiFramework, modifier -> Chart8(uiFramework, modifier) },
     { uiFramework, modifier -> Chart9(uiFramework, modifier) },
     { uiFramework, modifier -> Chart10(uiFramework, modifier) },
+    { uiFramework, modifier -> Chart11(uiFramework, modifier) },
   )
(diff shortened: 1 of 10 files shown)


Gowsky:
Hello, @cristianboicu01. Thank you for your contribution.

>It may not be the most elegant solution

We agree. For example:
- The gesture-related changes are significantly overcomplicated.
- Markers with `showMarkerOnTap = true` disappear whenever the chart is redrawn, not just on scroll. For example, if you tap to show the marker, and then a data update occurs, the marker disappears. How long the marker remains visible is accidental.
- The intention was for markers with `showMarkerOnTap = true` to disappear on scroll, but that shouldn’t occur. They should move together with their targets. You seem to have worked around this in `Chart11` by using `persistentMarkers`.
- Pixels are used instead of dp.
- Magic numbers are used.
- The naming is unintuitive. “Tap” and “touch” don’t distinguish between the two kinds of triggering gestures sufficiently clearly.
- The new `CartesianMarkerVisibilityListener` function is unnecessary.
- `displayOnTap` shouldn’t be in `CartesianMarker` but rather in `CartesianChart`, which is what draws markers. `CartesianMarker` itself shouldn’t be concerned with when it is drawn.
- `Chart11` misuses the API in several ways. For example, a new `LineComponent` is unnecessarily created on every `ColumnProvider.getColumn` call, even though there are only two distinct `LineComponent`s.

For these reasons we cannot accept this PR.

PR #925, opened 2024-10-18
Adds line height parameter to text component

cristianboicu01:
Provides a way to customise the line height for a text component.


Gowsky (changes requested):
Hello, @cristianboicu01. Thank you for your contribution. The PR looks good overall, but we'd like to request 2 changes:
- Add a missing documentation for `lineHeight` and `lineHeightSp` parameters.
- Add an XML attribute for the line height.

cristianboicu01:
Hello @Gowsky. Thank you for the comments. 

Hope I got it right with the xml attributes. Let me know if I missed anything.

Gowsky:
>Let me know if I missed anything.

Adding an XML attribute declaration is not enough. You have to read it and apply it to `TextComponent`, as it won't work just like that.

vico/views/src/main/java/com/patrykandpatrick/vico/views/common/theme/TypedArray.kt
```diff
@@ -49,6 +49,19 @@ internal fun TypedArray.getRawDimension(
     }
   }
 
+internal fun TypedArray.getRawDimensionOrNull(
```
Gowsky, line 52: Please update `TypedArray.getRawDimension` to use this, to avoid code duplication.

vico/core/src/main/java/com/patrykandpatrick/vico/core/common/component/TextComponent.kt
```diff
@@ -446,6 +453,14 @@ public open class TextComponent(
     return result
   }
 
+  private fun computeSpacingAddition(context: MeasuringContext): Float {
+    return with(context) {
+      lineHeightSp?.let {
+        spToPx(lineHeightSp) - textPaint.defaultLineHeight
+      } ?: 0f
+    }
+  }
```
Gowsky, line 462:
```suggestion
  private fun MeasuringContext.computeSpacingAddition(): Float =
    if (lineHeightSp != null) {
      spToPx(lineHeightSp) - textPaint.defaultLineHeight
    } else {
      0f
    }
```

vico/core/src/main/java/com/patrykandpatrick/vico/core/common/component/TextComponent.kt
```diff
@@ -371,14 +375,15 @@ public open class TextComponent(
...
+      val spacingAddition = computeSpacingAddition(context)
 
       val correctedWidth =
         (when {
             rotationDegrees % 1f.piRad == 0f -> widthWithoutMargins
             rotationDegrees % 0.5f.piRad == 0f -> heightWithoutMargins
             else -> {
               val cumulatedHeight =
-                lineCount * textPaint.lineHeight + padding.verticalDp.wholePixels
+                lineCount * textPaint.defaultLineHeight + spacingAddition + padding.verticalDp.wholePixels
```
Gowsky, line 386: Does it work with multi-line?

Gowsky (changes requested):

vico/core/src/main/java/com/patrykandpatrick/vico/core/common/component/TextComponent.kt
cristianboicu01: At first I thought I missed the paranthesis here but adding them makes things worse.

Increasing the value of `cumulatedHeight` leads to a smaller `correctedWidth` value which in consequence decreases the width of the layout and text does not fit. 

I'm wondering if this `correctedWidth` value is needed at all. If so, can you provide a bit of context here?

Gowsky: > Increasing the value of cumulatedHeight leads to a smaller correctedWidth value which in consequence decreases the width of the layout

This is generally correct for rotated text (except when the angle is 90 or 270 degrees). If a specific maximum width (which is independent of rotation) is supposed to be respected, then as the text height increases, its width must decrease.

> I'm wondering if this correctedWidth value is needed at all.

Yes, it's needed for the reason mentioned above.

However, your approach increases `cumulatedHeight` by too much. Adding parentheses causes `spacingAddition` to be multiplied by `lineCount`, when it should actually be multiplied by `lineCount - 1`.

A bigger issue is that the property is called `lineHeight`, but it actually changes only the line spacing. The distance between the first line and the top of the `TextComponent` always remains the same. This should be corrected. We'll update the PR accordingly.

We'll update the PR and make all the necessary fixes.

Gowsky (approved):

Gowsky:
Thank you for your contribution @cristianboicu01!

PR #934, opened 2024-10-26
Add entryIndex to PointProvider.getPoint

Tyler-Lopez:
### Background

Broadly speaking, I would love to get more information in the functions which provide `CartesianLayerModel.Entry`.

The information I am most interested in is:
* The index of the entry within its `series`.
* The x-value in pixels which corresponds to the `CartesianLayerModel.Entry.x`.
* The y-value in pixels which corresponds to the `CartesianLayerModel.Entry.y`.

I was a little surprised to find that `seriesIndex` was not the index of the `CartesianLayerModel.Entry` within its `series` but more broadly the whole `series` index among other `series`.

This commit adds `entryIndex` to `PointProvider.getPoint` which behaves as I had expected `seriesIndex` to.

I think this value may be beneficial to add to other surfaces, like the marker visibility listener.

### Motivation

My implementation which brought the desire for these changes is a little particular.

I will first emphasize: the change of this commit isn't absolutely required for me, however, it is a nice optimization.

The other two things I am interested in which are not included in this commit: the x and y-value in pixels of the entry do feel like some change would be required.

In my implementation, I abstract the Vico implementation of a Cartesian chart behind a Composable which takes a generic `T: Number`. Of course, Vico uses simply `Double`, which is fine - however, I wanted my Composable to be able to use just any `Number`.

I use `PointProvider.getPoint` to associate the entry with a `LineCartesianLayer.Point` with what is effectively a `Map<T, LineCartesianLayer.Point>`. However, as the `entry.x` value is a `Double`, it cannot be used in this generic purpose without an unchecked cast. My implementation work-around here is to look at my series' x-values, and find the closest `T` to `entry.x`. However, if I had an integer `entryIndex`, I could use this to more efficiently look-up the correct `T` corresponding to the `entry`.

### Key Changes
* `forEachIn` to `forEachInIndexed`.
* Add `entryIndex` to `PointProvider.getPoint`.

### Alternatives

As previously mentioned, this change is not explicitly necessary for me - though it would make things easier.

I am also curious if it would be of interest to more broadly have something like an `EntryMetadata` object which could encapsulate, among other things, these 3 additions (entry index, x & y pixel coordinates of the entry) I have mentioned.

### DIFF
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianLayerModel.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianLayerModel.kt
@@ -90,10 +90,10 @@ internal fun List<CartesianLayerModel.Entry>.getXDeltaGcd(): Double {
   } ?: 1.0
 }
 
-internal inline fun <T : CartesianLayerModel.Entry> List<T>.forEachIn(
+internal inline fun <T : CartesianLayerModel.Entry> List<T>.forEachInIndexed(
   range: ClosedFloatingPointRange<Double>,
   padding: Int = 0,
-  action: (T, T?) -> Unit,
+  action: (index: Int, entry: T, nextEntry: T?) -> Unit,
 ) {
   var start = 0
   var end = 0
@@ -106,5 +106,5 @@ internal inline fun <T : CartesianLayerModel.Entry> List<T>.forEachIn(
   }
   start = (start - padding).coerceAtLeast(0)
   end = (end + padding).coerceAtMost(lastIndex)
-  (start..end).forEach { action(this[it], getOrNull(it + 1)) }
+  (start..end).forEach { action(it, this[it], getOrNull(it + 1)) }
 }
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/CandlestickCartesianLayer.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/CandlestickCartesianLayer.kt
@@ -29,7 +29,7 @@ import com.patrykandpatrick.vico.core.cartesian.data.CandlestickCartesianLayerMo
 import com.patrykandpatrick.vico.core.cartesian.data.CartesianChartRanges
 import com.patrykandpatrick.vico.core.cartesian.data.CartesianLayerRangeProvider
 import com.patrykandpatrick.vico.core.cartesian.data.MutableCartesianChartRanges
-import com.patrykandpatrick.vico.core.cartesian.data.forEachIn
+import com.patrykandpatrick.vico.core.cartesian.data.forEachInIndexed
 import com.patrykandpatrick.vico.core.cartesian.layer.CandlestickCartesianLayer.Candle
 import com.patrykandpatrick.vico.core.cartesian.marker.CandlestickCartesianLayerMarkerTarget
 import com.patrykandpatrick.vico.core.cartesian.marker.CartesianMarker
@@ -151,7 +151,7 @@ protected constructor(
     var candle: Candle
     val minBodyHeight = minCandleBodyHeightDp.pixels
 
-    model.series.forEachIn(ranges.minX..ranges.maxX) { entry, _ ->
+    model.series.forEachInIndexed(ranges.minX..ranges.maxX) { _, entry, _ ->
       candle = candles.getCandle(entry, model.extraStore)
       val candleInfo = drawingModel?.entries?.get(entry.x) ?: entry.toCandleInfo(yRange)
       val xSpacingMultiplier = ((entry.x - ranges.minX) / ranges.xStep).toFloat()
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/ColumnCartesianLayer.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/ColumnCartesianLayer.kt
@@ -29,7 +29,7 @@ import com.patrykandpatrick.vico.core.cartesian.data.CartesianValueFormatter
 import com.patrykandpatrick.vico.core.cartesian.data.ColumnCartesianLayerDrawingModel
 import com.patrykandpatrick.vico.core.cartesian.data.ColumnCartesianLayerModel
 import com.patrykandpatrick.vico.core.cartesian.data.MutableCartesianChartRanges
-import com.patrykandpatrick.vico.core.cartesian.data.forEachIn
+import com.patrykandpatrick.vico.core.cartesian.data.forEachInIndexed
 import com.patrykandpatrick.vico.core.cartesian.marker.CartesianMarker
 import com.patrykandpatrick.vico.core.cartesian.marker.ColumnCartesianLayerMarkerTarget
 import com.patrykandpatrick.vico.core.cartesian.marker.MutableColumnCartesianLayerMarkerTarget
@@ -160,7 +160,7 @@ protected constructor(
     model.series.forEachIndexed { index, entryCollection ->
       drawingStart = getDrawingStart(index, model.series.size, mergeMode) - scroll
 
-      entryCollection.forEachIn(ranges.minX..ranges.maxX) { entry, _ ->
+      entryCollection.forEachInIndexed(ranges.minX..ranges.maxX) { _, entry, _ ->
         val columnInfo = drawingModel?.getOrNull(index)?.get(entry.x)
         height =
           (columnInfo?.height ?: (abs(entry.y) / yRange.length)).toFloat() * layerBounds.height()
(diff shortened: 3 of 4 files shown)


Gowsky:
>I use PointProvider.getPoint to associate the entry with a LineCartesianLayer.Point with what is effectively a `Map<T, LineCartesianLayer.Point>`.

Could you share some details on what kind of behavior you’re implementing by these means? Are you switching between different `Point`s based on _y_ value, for example?

Tyler-Lopez:
Sure thing!

https://gist.github.com/Tyler-Lopez/78873fd54ea9bbb5b46df13072e52c57

This is a simplified example that demonstrates how I am switching between different `LineCartesianLayer.Point` given an `entry` and how it could be made more simple with an `entryIndex`

Tyler-Lopez:
Here is a visual demonstration: particularly the state that is set after I click _"Random distribution of many types"_

https://github.com/user-attachments/assets/1a64941c-ee96-48de-bdf0-7cea8128f286

Tyler-Lopez:
The abstraction of `Double` behind `<T: Number>` is not needed. It was added early as something that might benefit the developer ergonomics of the layer.

The slider comparison you draw is a great one and I hear you there. As a devil's advocate, I do think there is an argument to be made that if clients didn't have to convert their `Double` to `Float` and then back to a `Double` on value change that their usage of the slider would be more readable. That same friction applies here. That said, in components I have that wrap a slider for other reasons, I do not currently generalize the `Float`.

What is necessary concerning the abstraction layer, and this was probably not what you were asking, is that I want to hide the implementation details of Vico entirely from the usages of the abstracted layer. This is demonstrated in the reference launcher app I include in my comments where I can choose which implementation layer the abstraction layer uses under the hood.

More generally, I am curious where your thoughts are on providing more contextual information (entry index and the canvas values previously mentioned) here. I recognize this particular one, the entry index on the point provider, is not a required change for functionality on my end as there are several options, including what I am already doing with `minBy`.

Gowsky:
> What _is_ necessary concerning the abstraction layer is that I want to hide the implementation details of Vico entirely from the usages of the abstracted layer. If a different, non-Vico Cartesian chart implementation was swapped-in which used `Float`, I would want the wrapper to not need to change to reflect that.

No API is made for such usage. If instead of using an API as designed, you try to force it into an arbitrary predefined box, complications are certain. There's no single contract for chart libraries. Different APIs offer different functionality and therefore must be interacted with differently. Each library has its own set of features, and some of them may be unique. Such abstraction would need to use a very basic set of features that are common among all other chart libraries you might use. Otherwise, the abstraction would need to be changed, and a simple library swap wouldn't be possible defying the purpose of the abstraction. 

On the one hand, the generalization to `Number` is excessive. On the other, forcing everything into a composable is too restrictive — it makes it impossible to store the `CartesianChartModelProducer` in the `ViewModel` and interact with it there, which is recommended.

Abstraction involves identifying and extracting common parts, not forcing all implementations into a single form that doesn't truly match any of them. A wrapper that works with many chart libraries is of little use if it supports none of them well and requires workarounds. As already mentioned, when switching to a new implementation, you may make changes in only one place, but making them will be twice as hard as it should be due to the inability to use the new API as designed.

> However, I do think there is an argument to be made that if clients didn't have to convert their `Double` to `Float` and then back to a `Double` on value change that their usages of `Slider` would be more readable.

Possibly, but that's off-topic. The question was whether you create similar wrappers for other components when they do use a specific type.

> where I wrap a `Slider` for other reasons, I do not currently generalize the `Float`

Similarly, you probably don't have generic wrappers for most larger, more comparable APIs, which are specific not only in terms of data types but also broader usage patterns. If you tried to create such wrappers, the story would be similar.

> More generally, I am curious where your thoughts are on providing more contextual information (entry index and the canvas values previously mentioned) here.

We're open to exposing such information if it's helpful under normal usage. We can't add parameters only to enable workarounds. Do you have any other use case in mind for `entryIndex`? What's the use case for the canvas coordinates? We don't want to pollute the API with unnecessary function parameters.

Tyler-Lopez:
I do disagree lightly with the overarching comments around the wrapper. I don't think it is productive to go into that further too much here, though I appreciate the perspective and will continue to consider your thoughts as I move forward.

Regarding `entryIndex`, I do not have a required usage. I might float that this would facilitate showing only every second or third point in a way that `entry.x` does not. That said: I do not need that currently, and I hear you on concerns regarding pollution of the API. Feel free to close this PR if that is not of interest to you.

On the canvas coordinates, now that I look at `CartesianMarkerVisibilityListener`, Vico may already support my intention nearly completely if `canvasX` is the x pixel coordinate of the entire `CartesianChartHost`. I'll add more context below, but unless you are curious, please do not feel the need to comment; I will investigate this on my end.

In this example of the Apple Health steps count chart, notice how the first screenshot shows textual content above the chart. In the second screenshot, the textual content disappears with this floating label clearly above (and not part of) the chart. Candidly, I do not care for this design. However, I expect that a similar design will become a product requirement for me.

The requirement here, as I see it, is to be able to take an x-value of the chart and know where it is drawn so that we might be able to associate it with a different Composable outside of the chart.

As `canvasX` is exposed, if that lines up as expected with the x-pixel of the `CartesianChartHost`, we could correlate chart content outside of the chart. It may be required for `canvasY` to be exposed, if this is available, but I have not yet explored this. I'll get back to you on this later.

Tyler-Lopez:
As a quick update, Vico does indeed already excellently support that with `canvasX` exposed in `CartesianMarkerVisibilityListener` via. `CartesianMarker.Target`, which I had missed before I requested this. Here, the red line is drawn outside of `CartesianChartHost`. I do believe I would still like to see `canvasY` exposed if that would be possible. The rationale would be that this might better facilitate positioning the depth of this red line into the chart - though I'll consider that more deeply before separately opening a feature request and potentially trying to contribute it myself.

https://github.com/user-attachments/assets/97b9419a-91a3-4321-9174-7186731913fb

Gowsky:
> I do disagree lightly with the overarching comments around the wrapper.

I think their accuracy is evident from the example. It's clear that Vico's API is used suboptimally in your code. It's also clear that this suboptimal use is made necessary by the wrapper. Thus, the wrapper inhibits correct usage of the implementation, which is an issue.

The only way to avoid this conclusion would be to suggest that the wrapper perfectly supports all correct chart APIs, and if there's friction with Vico, then it's due to an objective flaw in Vico's API. However, no such flaw has been demonstrated, and it can't be — there's nothing incorrect (or even unusual) about the use of `Double` or the presence of a class that's best used in `ViewModel`s.

You're welcome to use Vico however you wish. You may determine that the wrapper's disadvantages and the resulting suboptimal use of Vico aren't deal-breakers in your particular case. However, we can't accept a change when the rationale is based on overall faulty usage. APIs aren't designed to all be forced into the exact same box.

> Regarding `entryIndex`, I do not have a required usage. I might float that this would facilitate showing only every second or third point in a way that `entry.x` does not.

This is easy to achieve with extras, but it's still a valid reason to expose such data, so it could be considered. However, for consistency, entry indices would then have to be exposed in all similar APIs, not just `getPoint`. It would probably be best to add an `index` property to `Entry`.

Since the changes included in this PR only serve the initially described purpose, I unfortunately have to close it. As described above, we can't pollute the API with parameters that serve solely to accomodate flawed usage patterns. When used normally, Vico's API directly supports your use case of styling points based on _x_ value.

> I do believe I would still like to see `canvasY` exposed if that would be possible.

It already is, but it's in the `Target` implementations, so you access it slightly differently. See e.g. `LineCartesianLayerMarkerTarget.Point.canvasY`. Also see [here](https://patrykandpatrick.com/vico/wiki/cartesian-charts/markers#cartesianmarkertarget).

Tyler-Lopez:
@Gowsky Would you like me to open a feature request for adding an index property to `Entry`? Even if I don't use it, I would be excited to contribute to adding that if you would like and could open a separate PR towards that effort. Alternatively, if you would prefer for the feature addition to be proceeded by a genuine request for it to avoid unnecessary pollution I totally understand that as well.

Gowsky:
> Would you like me to open a feature request for adding an index property to `Entry`? Even if I don't use it, I would be excited to contribute to adding that if you would like and could open a separate PR towards that effort.

Please feel free to open a feature request and mention that you're open to contributing. We'd still have to evaluate this more thoroughly in terms of API design.

Regarding your code, instead of the `minByOrNull` workaround, you can use a map linking `Double` _x_ values to `T` _x_ values. This will offer significantly better performance, and it will be more concise. (In fact, it will be about as compact as the index-based solution.)

```kt
val dataXValuesKey = remember { ExtraStore.Key<Map<Double, T>>() }
```

```kt
extraStore[dataXValuesKey] = dataX.associateBy { it.toDouble() }
```

```kt
val xValue = dataXValues.getValue(entry.x)
```

Please also note that you should be using `ExtraStore.get` (`extraStore[...]`), not `ExtraStore.getOrNull`. This makes `getPoint` even shorter:

```kt
override fun getPoint(
    entry: LineCartesianLayerModel.Entry,
    seriesIndex: Int,
    extraStore: ExtraStore,
): LineCartesianLayer.Point? {
    val pointColor =
        extraStore[pointProviderKey]
            .getPointColorAtXValue(extraStore[dataXValuesKey].getValue(entry.x))
    ...
}
```

Finally, the `withContext(Dispatchers.Main)` in the `LaunchedEffect` is redundant. `Transaction`s should be processed off the main thread, and they're moved to `Dispatchers.Default` internally regardless of what you do outside.

In the case of the existing code, there's no need to use `withContext` at all. For the map solution above, it would be a good idea to use `withContext(Dispatchers.Default)`, so the map is created off the main thread.

PR #940, opened 2024-10-31
Manually provide vertical insets to LineCartesianLayer

Tyler-Lopez:
WIP PR which is in response to: https://github.com/patrykandpatrick/vico/discussions/939


vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/LineCartesianLayer.kt
```diff
@@ -575,8 +575,8 @@ protected constructor(
         xSpacing = xSpacing,
         scalableStartPadding = layerPadding.scalableStartDp.pixels,
         scalableEndPadding = layerPadding.scalableEndDp.pixels,
-        unscalableStartPadding = maxPointSize.half + layerPadding.unscalableStartDp.pixels,
-        unscalableEndPadding = maxPointSize.half + layerPadding.unscalableEndDp.pixels,
```
Tyler-Lopez, line 579: I'm wondering if there is anything lost in not using `ExtraStore` any longer to derive even the previous change to `CartesianLayerPadding` and if it is worse to commit to that further like this PR would.

Tyler-Lopez: I'm very crudely exploring an extrastore exposing alternative, where we take the "Provider" paradigm here

Gowsky:
Closing due to inactivity.

PR #941, opened 2024-11-01
Update layerPadding to getLayerPadding w/ ExtraStore

Tyler-Lopez:
### Background

This commit introduces CartesianLayerPaddingProvider, a means to retrieve CartesianLayerPadding with an `ExtraStore`. This commit does not substantially conflict with pre-existing functionality as `CartesianLayerPaddingProvider.fixed` is identical.

### Motivation

`CartesianLayerPadding` was recently added. This allows clients to provide padding to the start and end of a chart. However, these values are provided without an `ExtraStore`. Therefore, I believe that the values provided will never be synchronously safe.

With an `ExtraStore`, we can calculate the correct spacing safely.

### Future Work

The discovery of the desire to change this parameter came from some advanced requirements with points. 

https://github.com/patrykandpatrick/vico/discussions/939

I would like to move the horizontal and vertical insets achieved through `getLargestPoint` to being client-provided in `CartesianLayerPadding`. The basic demand for that is that currently, `getLargestPoint` + `unscalableStartPadding` can result in an unnecessary amount of padding. Further, imagine that we only had a point at the very middle of the graph. We should, in that case, not inset the chart at all to account for the point horizontally. We should do the top if it is the maximum value in the series.

So, I would also like to introduce `unscalableTopPadding` and `unscalableBottomPadding` and change `getLargestPoint` to only impact the point spacing.

### Note

I could be completely incorrect in the basic assertion here that this is necessary to correctly have synchronous padding: for example, to avoid a bad state if the padding, which is dependent on the line thickness, changes before the transaction finishes. Please correct me if I am wrong.

### DIFF
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChart.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChart.kt
@@ -52,7 +52,7 @@ public fun rememberCartesianChart(
   bottomAxis: Axis<Axis.Position.Horizontal.Bottom>? = null,
   marker: CartesianMarker? = null,
   markerVisibilityListener: CartesianMarkerVisibilityListener? = null,
-  layerPadding: CartesianLayerPadding = cartesianLayerPadding(),
+  layerPadding: ((ExtraStore) -> CartesianLayerPadding) = { cartesianLayerPadding() },
   legend: Legend<CartesianMeasuringContext, CartesianDrawingContext>? = null,
   fadingEdges: FadingEdges? = null,
   decorations: List<Decoration> = emptyList(),
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
@@ -146,7 +146,8 @@ internal fun CartesianChartHostImpl(
       ranges = ranges,
       scrollEnabled = scrollState.scrollEnabled,
       zoomEnabled = scrollState.scrollEnabled && zoomState.zoomEnabled,
-      layerPadding = chart.layerPadding,
+      layerPadding =
+        remember(chart.layerPadding, model.extraStore) { chart.layerPadding(model.extraStore) },
       spToPx = with(LocalContext.current) { ::spToPx },
     )
 
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/CartesianChart.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/CartesianChart.kt
@@ -62,7 +62,7 @@ import kotlin.math.abs
  * @property layers the [CartesianLayer]s.
  * @property marker appears when the [CartesianChart] is tapped.
  * @property markerVisibilityListener allows for listening to [marker] visibility changes.
- * @property layerPadding stores the [CartesianLayer] padding values.
+ * @property layerPadding returns the [CartesianLayerPadding].
  * @property legend the legend.
  * @property fadingEdges applies a horizontal fade to the edges of the [CartesianChart], provided
  *   that it’s scrollable.
@@ -80,7 +80,7 @@ public open class CartesianChart(
   @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) public val marker: CartesianMarker? = null,
   protected val markerVisibilityListener: CartesianMarkerVisibilityListener? = null,
   @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-  public val layerPadding: CartesianLayerPadding = CartesianLayerPadding(),
+  public val layerPadding: ((ExtraStore) -> CartesianLayerPadding) = { CartesianLayerPadding() },
   protected val legend: Legend<CartesianMeasuringContext, CartesianDrawingContext>? = null,
   protected val fadingEdges: FadingEdges? = null,
   protected val decorations: List<Decoration> = emptyList(),
@@ -408,7 +408,7 @@ public open class CartesianChart(
     bottomAxis: Axis<Axis.Position.Horizontal.Bottom>? = this.bottomAxis,
     marker: CartesianMarker? = this.marker,
     markerVisibilityListener: CartesianMarkerVisibilityListener? = this.markerVisibilityListener,
-    layerPadding: CartesianLayerPadding = this.layerPadding,
+    layerPadding: ((ExtraStore) -> CartesianLayerPadding) = this.layerPadding,
     legend: Legend<CartesianMeasuringContext, CartesianDrawingContext>? = this.legend,
     fadingEdges: FadingEdges? = this.fadingEdges,
     decorations: List<Decoration> = this.decorations,
--- a/vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
+++ b/vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
@@ -47,6 +47,7 @@ import com.patrykandpatrick.vico.views.common.gesture.ChartScaleGestureListener
 import com.patrykandpatrick.vico.views.common.gesture.MotionEventHandler
 import com.patrykandpatrick.vico.views.common.isLtr
 import com.patrykandpatrick.vico.views.common.theme.ThemeHandler
+import java.util.Objects
 import kotlin.math.abs
 import kotlin.properties.Delegates.observable
 import kotlin.properties.ReadWriteProperty
@@ -81,7 +82,7 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
       ranges = CartesianChartRanges.Empty,
       scrollEnabled = false,
       zoomEnabled = false,
-      layerPadding = themeHandler.chart?.layerPadding ?: CartesianLayerPadding(),
+      layerPadding = CartesianLayerPadding(),
       spToPx = context::spToPx,
     )
 
@@ -95,6 +96,8 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
   private var scrollDirectionResolved = false
 
   private var horizontalDimensions = MutableHorizontalDimensions()
+
+  private var previousLayerPaddingHashCode: Int? = null
 
   /**
    * Houses information on the [CartesianChart]’s scroll value. Allows for scroll customization and
@@ -128,7 +131,6 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
   /** The [CartesianChart] displayed by this [View]. */
   public var chart: CartesianChart? by
     observable(themeHandler.chart) { _, _, newValue ->
-      if (newValue != null) measuringContext.layerPadding = newValue.layerPadding
       tryInvalidate(chart = newValue, model = model, updateRanges = true)
     }
 
@@ -243,10 +245,17 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
     updateRanges: Boolean,
   ) {
     measuringContext.model = model ?: return
-    if (chart != null && updateRanges) {
-      ranges.reset()
-      chart.updateRanges(ranges, model)
-      measuringContext.ranges = ranges.toImmutable()
+    if (chart != null) {
+      val layerPaddingHashCode = Objects.hash(chart.layerPadding, model.extraStore)
+      if (layerPaddingHashCode != previousLayerPaddingHashCode) {
+        measuringContext.layerPadding = chart.layerPadding(model.extraStore)
+        previousLayerPaddingHashCode = layerPaddingHashCode
+      }
+      if (updateRanges) {
+        ranges.reset()
+        chart.updateRanges(ranges, model)
+        measuringContext.ranges = ranges.toImmutable()
+      }
     }
     if (isAttachedToWindow) invalidate()
   }
--- a/vico/views/src/main/java/com/patrykandpatrick/vico/views/common/theme/ThemeHandler.kt
+++ b/vico/views/src/main/java/com/patrykandpatrick/vico/views/common/theme/ThemeHandler.kt
@@ -211,6 +211,7 @@ internal class ThemeHandler(private val context: Context, attrs: AttributeSet?) 
     val candlestickLayer =
       if (layerFlags.hasFlag(CANDLESTICK_LAYER)) typedArray.getCandlestickCartesianLayer(context)
       else null
+    val layerPadding = typedArray.getLayerPadding()
 
     return CartesianChart(
       layers =
@@ -225,7 +226,7 @@ internal class ThemeHandler(private val context: Context, attrs: AttributeSet?) 
       endAxis = typedArray.getAxis(Axis.Position.Vertical.End),
       bottomAxis = typedArray.getAxis(Axis.Position.Horizontal.Bottom),
       fadingEdges = typedArray.getFadingEdges(),
-      layerPadding = typedArray.getLayerPadding(),
+      layerPadding = { layerPadding },
     )
   }
 
--- a/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/charts/Chart1.kt
+++ b/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/charts/Chart1.kt
@@ -83,7 +83,7 @@ private fun ComposeChart1(modelProducer: CartesianChartModelProducer, modifier: 
             itemPlacer = remember { HorizontalAxis.ItemPlacer.segmented() },
           ),
         marker = marker,
-        layerPadding = cartesianLayerPadding(scalableStart = 16.dp, scalableEnd = 16.dp),
+        layerPadding = { cartesianLayerPadding(scalableStart = 16.dp, scalableEnd = 16.dp) },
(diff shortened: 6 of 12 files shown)


Tyler-Lopez:
Hi @Gowsky!

Thanks so much for providing the context. This will take some time for me to chew over, but I do understand the ask to simply make this an `ExtraStore` providing lambda and I have made that change. Nice suggestion!

If there is anything else here I am missing, please let me know and I would be happy to adjust this.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChart.kt
```diff
@@ -52,7 +52,7 @@ public fun rememberCartesianChart(
   bottomAxis: Axis<Axis.Position.Horizontal.Bottom>? = null,
   marker: CartesianMarker? = null,
   markerVisibilityListener: CartesianMarkerVisibilityListener? = null,
-  layerPadding: CartesianLayerPadding = cartesianLayerPadding(),
+  getLayerPadding: ((ExtraStore) -> CartesianLayerPadding) =  { CartesianLayerPadding() },
```
Gowsky, line 55: Please use the compose factory function that was used before.
```suggestion
  getLayerPadding: ((ExtraStore) -> CartesianLayerPadding) =  { cartesianLayerPadding() },
```

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
```diff
@@ -146,7 +146,7 @@ internal fun CartesianChartHostImpl(
       ranges = ranges,
       scrollEnabled = scrollState.scrollEnabled,
       zoomEnabled = scrollState.scrollEnabled && zoomState.zoomEnabled,
-      layerPadding = chart.layerPadding,
+      layerPadding = chart.getLayerPadding.invoke(model.extraStore),
```
Gowsky, line 149: Please call the lambda normally. Also, the result should be cached until `ExtraStore` changes.
```suggestion
      layerPadding = remember(chart, chart.getLayerPadding, model.extraStore) { chart.getLayerPadding(model.extraStore) },
```

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/CartesianChart.kt
```diff
@@ -77,7 +77,7 @@ public open class CartesianChart(
   @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) public val marker: CartesianMarker? = null,
   protected val markerVisibilityListener: CartesianMarkerVisibilityListener? = null,
   @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-  public val layerPadding: CartesianLayerPadding = CartesianLayerPadding(),
+  public val getLayerPadding: ((ExtraStore) -> CartesianLayerPadding) = { CartesianLayerPadding() },
```
Gowsky, line 80: For simplicity, we don't use the `get` prefix for `ExtraStore` lambdas. They're essentially just scopes for expressions.

You can find a similar approach in other libraries, e.g. the Kotlin standard library:

```kt
inline fun check(
    value: Boolean, 
    lazyMessage: () -> Any
)
```

```suggestion
  public val layerPadding: ((ExtraStore) -> CartesianLayerPadding) = { CartesianLayerPadding() },
```

vico/views/src/main/java/com/patrykandpatrick/vico/views/common/theme/ThemeHandler.kt
```diff
@@ -227,7 +227,7 @@ internal class ThemeHandler(private val context: Context, attrs: AttributeSet?)
       endAxis = baseTypedArray.getAxis(Axis.Position.Vertical.End),
       bottomAxis = baseTypedArray.getAxis(Axis.Position.Horizontal.Bottom),
       fadingEdges = baseTypedArray.getFadingEdges(),
-      layerPadding = baseTypedArray.getLayerPadding(),
+      getLayerPadding = { baseTypedArray.getLayerPadding() },
```
Gowsky, line 230: `baseTypedArray.getLayerPadding()` must be called outside of the lambda, as the `baseTypedArray` is recycled after `getChart` function completes. The lambda should return just a property.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Gowsky, line 1: The lambda is not called when the `ExtraStore` changes.

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/CartesianChart.kt
```diff
@@ -59,7 +59,7 @@ import kotlin.math.abs
  * @property layers the [CartesianLayer]s.
  * @property marker appears when the [CartesianChart] is tapped.
  * @property markerVisibilityListener allows for listening to [marker] visibility changes.
- * @property layerPadding stores the [CartesianLayer] padding values.
+ * @property getLayerPadding retrieves the [CartesianLayer] padding values given the [ExtraStore].
```
Gowsky, line 62:
```suggestion
 * @property layerPadding returns the [CartesianLayerPadding].
```

Gowsky (changes requested):

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
Tyler-Lopez: Good call! I'm just curious, are we sure we need all of those keys? Particularly the `chart` one. I did add it, but just calling out perhaps only `layerPadding` and `extraStore` are necessary?

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Tyler-Lopez: The concern for you here being that, if `ExtraStore` changes, `measuringContext` does not have its `layerPadding` set as it does when the `chart` is updated, correct? Where would you like to see this? In `registerForUpdates`? Or, perhaps similar to how `chart` uses an `observable`?

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
Gowsky: >Particularly the chart one.

Yes, it's not needed, actually. There was a mistake in the suggestion.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Gowsky: >The concern for you here being that, if ExtraStore changes, measuringContext does not have its layerPadding set as it does when the chart is updated, correct?

Yes. That defeats the purpose of the lambda because it's not called when the extras change. 

>Where would you like to see this?

The lambda should be called in `tryInvalidate`. Caching should be added, just like in Compose.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -120,7 +120,6 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
   /** The [CartesianChart] displayed by this [View]. */
   public var chart: CartesianChart? by
     observable(themeHandler.chart) { _, _, newValue ->
-      if (newValue != null) measuringContext.layerPadding = newValue.layerPadding
```
Tyler-Lopez, line 123: This no longer seems necessary if we are setting the layer padding in `tryInvalidate` which is always invoked just below. One change is that I am updating the layer padding even if the incoming value is `null` now to the default `CartesianLayerPadding`

Tyler-Lopez: I did add a commit to add a `updateLayerPadding` parameter in an attempt to avoid unnecessary new layer padding fetching

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Gowsky: The lambda still isn't called when the extras change, as `updateLayerPadding` is `false` in `onModelCreated`. There's no need for the `updateLayerPadding` parameter. All we need is for the lambda to be called from `tryInvalidate`, but only when needed (when it or the extras change), just like in `CartesianChartHostImpl`. There's also no need to apply `CartesianLayerPadding()` when `chart` is null.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Gowsky: Please see the comment [here](https://github.com/patrykandpatrick/vico/pull/941/files#r1873930567).

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Tyler-Lopez: Got it - thank you for your patience and guidance here! 

So, the model updating will _always_ update the `ExtraStore`, and thus, in the lambda of `modelProducer?.registerForUpdates` when we invoke `setModel`, we can guarantee that we always need to update the `layerPadding` of the `measuringContext` and there is thus no need to ever have a `updateLayerPadding` parameter in `setModel`.

I am retaining `updateLayerPadding` in `tryInvalidate` for the purpose of only invoked the lambda when necessary: either when the lambda object is different when the `chart` object is updated or when the `model` is updated.

Gowsky: >the model updating will *always* update the ExtraStore

Not necessarily. It’s entirely possible for the series to be updated but for the extra values to remain the same.

>I am retaining updateLayerPadding

It’s not needed. It’s best to check for lambda and `ExtraStore` changes in `tryInvalidate`.  The lambda should be called when at least one of them changes compared to the previous `tryInvalidate` call.

Tyler-Lopez: Ahhh - sorry, I only have just understood what you were referring to with caching; this makes sense. I have added `lastDrawnExtraStore` and `lastDrawnLayerPadding` to serve this purpose. Of course, happy to move around these names as you would prefer.

Gowsky: @patrickmichalik will follow up with the final suggestions. The caching logic can be slightly improved, and the lambda is called with the incorrect `ExtraStore`, so it doesn't work, effectively.

Also, please ensure to reformat the code using ktfmt, as per the contribution guidelines.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -73,7 +74,7 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
       ranges = CartesianChartRanges.Empty,
       scrollEnabled = false,
       zoomEnabled = false,
-      layerPadding = themeHandler.chart?.layerPadding ?: CartesianLayerPadding(),
+      layerPadding = themeHandler.chart?.layerPadding?.invoke(extraStore) ?: CartesianLayerPadding(),
```
patrickmichalik, line 77: `CartesianChartModel.extraStore` isn’t available at this stage, but an initial value is required. `CartesianLayerPadding()` works; it’ll be replaced before drawing. (Down the road, it may be beneficial to introduce a `CartesianLayerPadding.Zero` singleton, but this needn’t be handled in this PR.)

```suggestion
      layerPadding = CartesianLayerPadding(),
```

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -88,6 +89,9 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
 
   private var horizontalDimensions = MutableHorizontalDimensions()
 
+  private var lastDrawnExtraStore: ExtraStore? = null
+  private var lastDrawnLayerPadding: ((ExtraStore) -> CartesianLayerPadding)? = null
```
patrickmichalik, line 93:
```suggestion
  private var previousLayerPaddingHashCode: Int? = null
```

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -240,12 +243,22 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
...
     measuringContext.model = model ?: return
     if (chart != null && updateRanges) {
       ranges.reset()
       chart.updateRanges(ranges, model)
       measuringContext.ranges = ranges.toImmutable()
     }
+
+    lastDrawnExtraStore = extraStore
+    lastDrawnLayerPadding = chart?.layerPadding
+
```
patrickmichalik, line 261:
```suggestion
    measuringContext.model = model ?: return
    if (chart != null) {
      val layerPaddingHashCode = Objects.hash(chart.layerPadding, model.extraStore)
      if (layerPaddingHashCode != previousLayerPaddingHashCode) {
        previousLayerPaddingHashCode = layerPaddingHashCode
        measuringContext.layerPadding = chart.layerPadding(model.extraStore)
        previousLayerPaddingHashCode = layerPaddingHashCode
      }
      if (updateRanges) {
        ranges.reset()
        chart.updateRanges(ranges, model)
        measuringContext.ranges = ranges.toImmutable()
      }
    }
```

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
```diff
@@ -146,7 +146,12 @@ internal fun CartesianChartHostImpl(
       ranges = ranges,
       scrollEnabled = scrollState.scrollEnabled,
       zoomEnabled = scrollState.scrollEnabled && zoomState.zoomEnabled,
-      layerPadding = chart.layerPadding,
+      layerPadding = remember(
+        chart.layerPadding,
+        model.extraStore,
+      ) {
+        chart.layerPadding(model.extraStore)
+      },
```
patrickmichalik, line 154: There’s a formatting issue here. `ktfmt` makes this change:

```suggestion
      layerPadding =
        remember(chart.layerPadding, model.extraStore) { chart.layerPadding(model.extraStore) },
```

patrickmichalik (changes requested):
Hello, and thanks for the PR! As per @Gowsky’s comment, I’d like to raise two matters, both connected to `CartesianChartView`.

The first one relates to determining when `chart.layerPadding` should be invoked in `tryInvalidate`. Rather than storing references to the previous lambda and `ExtraStore`, we can employ a solution based on hash codes. Such a system is already in place in `CartesianChart` for `persistentMarkers`, another `ExtraStore` lambda. It’s slightly more flexible.

Second—and more importantly—`CartesianChart.layerPadding` should be passed `CartesianChartModel.extraStore`, not `ChartView.extraStore`. The latter houses only extras specific to the host instance. Consumer-specified extras are added to `CartesianChartModelProducer.extraStore`, which is combined with `ChartView.extraStore` during `Transaction` processing. The resulting `ExtraStore` is then accessible via `CartesianChartModel.extraStore`.

The attached suggestions, which depend on one another, address both of these points. Please note that two import changes are required in addition. I’m unable to fully account for these via interactive suggestions; comments can be added only for lines in the vicinity of existing modifications.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
Tyler-Lopez: I personally was struggling to run `ktfmt` here. At least in such a way that resulted in only a small change. I got the `ktfmt` plug-in, enabled it, and attempted to format - but noticed quite a lot changing.

I did adopt this change, though.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
Gowsky: @Tyler-Lopez, it seems that you removed `if (isAttachedToWindow) invalidate()` by accident, so I fixed it.

Tyler-Lopez: Thanks so much for catching that discrepancy and fixing it!

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
Gowsky: `ktfmt` has multiple code styles. As mentioned in the contributing guidelines, we use the Google style. You probably forgot to set it. When using `ktfmt` directly (via the CLI), you apply it using `--google-style`. The plugin you used probably has a setting for the code style.

Gowsky (approved):

patrickmichalik (approved):
Cheers!

PR #992, opened 2025-01-07
CubicPointConnector - Remove Impossible Logic

Tyler-Lopez:
Recently, the constructor of `CubicPointConnector` was updated to produce a run-time crash when curvature is == 0.

This did not happen before, and as such, there is a piece of logic to handle the curvature == 0 case. This can now be removed.

### DIFF
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/CubicPointConnector.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/CubicPointConnector.kt
@@ -34,15 +34,11 @@ internal data class CubicPointConnector(private val curvature: Float) :
     x2: Float,
     y2: Float,
   ) {
-    if (curvature == 0f) {
-      path.lineTo(x2, y2)
-    } else {
-      val xDelta =
-        (Y_MULTIPLIER * abs(y2 - y1) / context.layerBounds.height()).coerceAtMost(1f) *
-          curvature *
-          (x2 - x1)
-      path.cubicTo(x1 + xDelta, y1, x2 - xDelta, y2, x2, y2)
-    }
+    val xDelta =
+      (Y_MULTIPLIER * abs(y2 - y1) / context.layerBounds.height()).coerceAtMost(1f) *
+        curvature *
+        (x2 - x1)
+    path.cubicTo(x1 + xDelta, y1, x2 - xDelta, y2, x2, y2)
   }
 
   private companion object {


vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/layer/CubicPointConnector.kt
```diff
@@ -34,15 +34,11 @@ internal data class CubicPointConnector(private val curvature: Float) :
     x2: Float,
     y2: Float,
   ) {
-    if (curvature == 0f) {
-      path.lineTo(x2, y2)
-    } else {
```
Tyler-Lopez, line 39: Personally, I preferred when the logic was `curvature >= 0` instead of `curvature > 0`.

Tyler-Lopez: I like the option to construct a "curved" line with 0 curvature as, if I wanted to possibly dynamically change the curvature as a client, I could do so quite easily simply by mutating the `curvature` field of `CubicPointConnector`. Now, there is more client logic required to do that as I would need to move this `curvature == 0f` check to my logic and alternatively construct with the `Sharp` singleton in that case.

Very minor, and no issue, but that was my thought.

Gowsky: > if I wanted to possibly dynamically change the curvature as a client

That seems unusual, but if you have an example of an actual use case where it's required, feel free to open a discussion in "Ideas."

> It also thus is a touch unintuitive, to me, that providing 0 here would RTE. I would think I should be able to make a curved line with 0% curvature.

There's a lint error for invalid arguments (including 0), so to get a runtime exception, you must first ignore the error, which you shouldn't do.

As per my previous comment, we added `Sharp` for clarity. We previously received some questions regarding how to create sharp lines.

Setting aside the use case with dynamic curvature changes, making `CubicPointConnector` accept `curvature = 0f` will always be slightly suboptimal:

- If the `if`-`else` is kept and `lineTo` is used for `curvature = 0f`, then a check will be run on every `connect` invocation (and thus possibly dozens of times per frame). It's preferable to avoid this (another reason why `Sharp` was added).
- Keeping just the logic that's currently under `else` and allowing `curvature = 0f` would work. In this case, the anchors would overlap the line points, giving a straight line. However, creating a straight line this way is slower than using `lineTo`, so this should also be avoided.

A better solution could be to update the `cubic` factory function so that it returns `Sharp` for `curvature = 0f`. This would avoid both of the issues mentioned above. However, for static curvature, this isn't really a tangible improvement, since you can simply use `Sharp`. In fact, `Sharp` would still be preferable, since it would be shorter and more direct, avoding the unnecessary function call.

Thus, I think making such a change is worthwhile only if there is a realistic use case for dynamically switching between sharp and cubic lines, since it would have a real benefit there. (It would avoid the extra client logic you mentioned.)

Tyler-Lopez: > If the if-else is kept and lineTo is used for curvature = 0f, then a check will be run on every connect invocation

Great point! I think, even beyond it being worthwhile if there is a realistic use case, it would have to be a very common use case - otherwise, individual cases of the client logic aren't a big deal. I'm convinced here; thanks for the thoughtful replies!

patrickmichalik (approved):
Hello! Thanks for the enhancement. This will be merged shortly.

PR #994, opened 2025-01-13
Add BelowPoint to LabelPosition

L-Andrade:
Hello! We have a use case in our app where we want to show a point's label below the point itself. Currently there are only two options related with the point: `AbovePoint` and `AroundPoint`. Adding `BelowPoint` adds more configurability to how we want labels to be shown.

I also had an idea for another improvement when modifying the code: `AbovePoint` (and this new `BelowPoint`) could be configured to either the topmost or bottommost value, instead of "hardcoded" to one or the other. I haven't added this because it would need additional changes and might require some additional discussion on how to achieve this behaviour

Here's a demo of how it looks like:

https://github.com/user-attachments/assets/8c001352-789d-438a-a7e0-8d8c2139fa3e

### DIFF
--- a/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/Marker.kt
+++ b/sample/src/main/java/com/patrykandpatrick/vico/sample/showcase/Marker.kt
@@ -126,7 +126,8 @@ internal fun rememberMarker(
           when (labelPosition) {
             LabelPosition.Top,
             LabelPosition.AbovePoint -> topMargin += label.getHeight(context) + tickSizeDp.pixels
-            LabelPosition.Bottom -> bottomMargin += label.getHeight(context) + tickSizeDp.pixels
+            LabelPosition.Bottom,
+            LabelPosition.BelowPoint -> bottomMargin += label.getHeight(context) + tickSizeDp.pixels
             LabelPosition.AroundPoint -> {}
           }
           layerMargins.ensureValuesAtLeast(top = topMargin, bottom = bottomMargin)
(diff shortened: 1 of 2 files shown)


Gowsky:
Hello, @L-Andrade. Thanks for your contribution. The code looks good, but it needs to be formatted using ktfmt as per the [contribution guidelines](https://www.patrykandpatrick.com/vico/guide/latest/contributing#pull-requests).

>I also had an idea for another improvement when modifying the code: AbovePoint (and this new BelowPoint) could be configured to either the topmost or bottommost value, instead of "hardcoded" to one or the other. I haven't added this because it would need additional changes and might require some additional discussion on how to achieve this behaviour

This can be achieved by making `LabelPosition` a `sealed class`. This would support the additional configuration via `AbovePoint` and `BelowPoint` constructors. Feel free to open a feature request or a PR.

L-Andrade:
> Hello, @L-Andrade. Thanks for your contribution. The code looks good, but it needs to be formatted using ktfmt as per the [contribution guidelines](https://www.patrykandpatrick.com/vico/guide/latest/contributing#pull-requests).
> 
> > I also had an idea for another improvement when modifying the code: AbovePoint (and this new BelowPoint) could be configured to either the topmost or bottommost value, instead of "hardcoded" to one or the other. I haven't added this because it would need additional changes and might require some additional discussion on how to achieve this behaviour
> 
> This can be achieved by making `LabelPosition` a `sealed class`. This would support the additional configuration via `AbovePoint` and `BelowPoint` constructors. Feel free to open a feature request or a PR.

Oh, sorry, missed that! I've pushed f664165c45ab97caf45f7a7507edf09e4c16044f to fix it.

For other improvements, I might open separate PRs in the future. Might I ask if a release can be built when this is merged? Would be cool to include it in our next app release. 

Thanks for checking!

Gowsky:
> Might I ask if a release can be built when this is merged? Would be cool to include it in our next app release.

Since Vico v2.0.0 has already been released, and this is a new feature, it will be included in Vico 2.1.0. We will release the first alpha soon.

patrickmichalik:
Hello, @L-Andrade! Thanks for the PR. [Vico 2.1.0 Alpha 1](https://github.com/patrykandpatrick/vico/releases/tag/v2.1.0-alpha.1), which includes these changes, is now available.

Issue #985, opened 2025-01-02
Chart in HorizontalPager consumes all swipes


mak1nt0sh:
### How to reproduce

Create simple line chart with marker inside either `HorizontalPager` & disable scroll via `scrollState = rememberVicoScrollState(scrollEnabled = false)`

### Observed behavior

Cannot scroll through values in `HorizontalPager`:

[ezgif-6-4096dc45a7.webm](https://github.com/user-attachments/assets/ffbc0303-bf3e-415a-a171-d0f510246e2b)

### Expected behavior

Same behaviour as in `2.0.0-beta.3` or provide a way to disable propagating the gestures when marker is visible.

### Vico version(s)

2.0.0-beta.5

### Android version(s)

Any

mak1nt0sh:
Also even when swiping left-right (horizontally) on the chart, vertical scroll somehow propagates to LazyColumn. In this example the TopAppBar starts collapsing.

[ezgif-6-e0744126a4.webm](https://github.com/user-attachments/assets/4a7442d5-7b94-4e53-be91-72fbefd76f54)

Gowsky:
Hello, @mak1nt0sh. This change was made in response to #942. Having charts entirely block pager scrolling was considered undesirable.

> or provide a way to disable propagating the gestures when marker is visible.

It looks like either behavior may be desirable, so such an option may be needed. Please open a feature request.

> Also even when swiping left-right (horizontally) on the chart in LazyColumn, vertical scroll somehow propagates to LazyColumn. In this example the TopAppBar starts collapsing.

This appears to be a separate topic. Please open a dedicated discussion.

MadestSamurai:
To avoid this temporarily, you could just add a drag gesture consumer in the chart's modifier.
For example: 
`modifier = Modifier.pointerInput(Unit) {
    detectDragGestures { change, _ ->
        change.consume()
    }
}`

Issue #942, opened 2024-11-03
Unable to scroll HorizontalPager when marker is set


shervinkoushan:
### How to reproduce

Create a chart with a marker. Put the chart in a HorizontalPager. 

### Observed behavior

Unable to scroll the pager as the marker consumes all touch events.

### Expected behavior

Possible to scroll the pager and highlight points by clicking on them.

### Vico version(s)

2.0.0-beta.2

### Android version(s)

14

Gowsky:
Hello, @shervinkoushan. Thank you for the report. It happens only for charts that aren't scrollable. We'll look into it.

shervinkoushan:
Hi @Gowsky. I am indeed able to scroll the `HorizontalPager` when I set `scrollEnabled` to true on the chart.
It would be nice if the marker was not shown when scrolling though.

Gowsky:
We plan on adding such an option. Please see [this comment](https://github.com/patrykandpatrick/vico/discussions/769#discussioncomment-10834453).

Gowsky:
[Vico v2 0.0-beta.4](https://github.com/patrykandpatrick/vico/releases/tag/v2.0.0-beta.4) fixes this.

PR #1006, opened 2025-01-30
Consume Move pointer events in `TouchEvent`

L-Andrade:
When a graph without horizontal scroll is in a horizontally scrollable component, such as a HorizontalPager, the events are not being consumed, resulting in a scroll when the user is interacting with the graph.
The `TouchEvent` should consume the move events so that the events are not processed by other scroll/touch handlers.

This change is only applied when scrolling is not enabled. When scrolling is enabled, the behaviour is the same.

Here's a before (old) and after (new) video with the change.

https://github.com/user-attachments/assets/030d09d6-e43f-4c7b-8e77-14dccee4d53b

https://github.com/user-attachments/assets/700e2d49-a603-4de5-80de-08a9c48c08e0

Please let me know if I missed anything, as this might be on purpose.

### DIFF
--- a/vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
+++ b/vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
@@ -127,10 +127,17 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
   private val motionEventHandler =
     MotionEventHandler(
       scroller = scroller,
+      consumeMoveEvents = themeHandler.consumeMoveEvents,
       density = resources.displayMetrics.density,
       onTouchPoint = ::handleTouchEvent,
       requestInvalidate = ::invalidate,
     )
+
+  /**
+   * Manages move touch events consumption by the view when scrolling is disabled and
+   * [CartesianChart.marker] is not null.
+   */
+  public var consumeMoveEvents: Boolean by motionEventHandler::consumeMoveEvents
 
   /** The [CartesianChart] displayed by this [View]. */
   public var chart: CartesianChart? by
@@ -286,20 +293,29 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
         false
       }
     val touchHandled = motionEventHandler.handleMotionEvent(event, scrollHandler)
+    val hasMarker = chart?.marker != null
+    when {
+      consumeMoveEvents && !scrollHandler.scrollEnabled && hasMarker -> {
+        parent.requestDisallowInterceptTouchEvent(true)
+      }
 
-    if (scrollDirectionResolved.not() && event.historySize > 0) {
-      scrollDirectionResolved = true
-      parent.requestDisallowInterceptTouchEvent(
-        event.movedXDistance > event.movedYDistance || event.pointerCount > 1
-      )
-    } else if (
-      event.actionMasked == MotionEvent.ACTION_UP || event.actionMasked == MotionEvent.ACTION_CANCEL
-    ) {
-      scrollDirectionResolved = false
+      scrollDirectionResolved.not() && event.historySize > 0 -> {
+        scrollDirectionResolved = true
+        parent.requestDisallowInterceptTouchEvent(
+          event.movedXDistance > event.movedYDistance || event.pointerCount > 1
+        )
+      }
+
+      event.isUpOrCancel() -> {
+        scrollDirectionResolved = false
+      }
     }
 
     return touchHandled || scaleHandled || superHandled
   }
+
+  private fun MotionEvent.isUpOrCancel(): Boolean =
+    actionMasked == MotionEvent.ACTION_UP || actionMasked == MotionEvent.ACTION_CANCEL
 
   private fun handleZoom(focusX: Float, zoomChange: Float) {
     val chart = chart ?: return
--- a/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
+++ b/vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
@@ -70,6 +70,8 @@ import kotlinx.coroutines.launch
  * @param animationSpec the [AnimationSpec] for difference animations.
  * @param animateIn whether to run an initial animation when the [CartesianChartHost] enters
  *   composition. The animation is skipped for previews.
+ * @param consumeMoveEvents whether move touch events will be consumed by the composable when
+ *   scrolling is disabled and [CartesianChart.marker] is not null.
  * @param placeholder shown when no [CartesianChartModel] is available.
  */
 @Composable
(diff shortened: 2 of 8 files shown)


Gowsky:
Hello, @L-Andrade. This has been the subject of several threads. Previously, gestures were consumed in this case. This behavior was [reported as a bug](https://github.com/patrykandpatrick/vico/issues/942) and replaced with the one that's currently in place. After this replacement, the new behavior was also [reported as a bug](https://github.com/patrykandpatrick/vico/issues/985).

Since either behavior has its fans, it seems that instead of changing the consumption logic once again, we should make this customizable, as mentioned [here](https://github.com/patrykandpatrick/vico/issues/985#issuecomment-2568280069). Are you interested in adding such a parameter? (It would have to be added for all three frameworks: Jetpack Compose, Compose Multiplatform, and views.)

In the meantime, [this workaround](https://github.com/patrykandpatrick/vico/issues/985#issuecomment-2570332828) may be helpful.

L-Andrade:
Hey @Gowsky 

Thanks for the reply! I'm definitely interested in doing so. 

I'll take a look when I have the time, not sure when that will be.

I'll tag you here if I have questions regarding the implementation.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/TouchEvent.kt
```diff
@@ -27,15 +27,14 @@ import com.patrykandpatrick.vico.core.common.Point
 
 internal fun Modifier.chartTouchEvent(
   setTouchPoint: ((Point?) -> Unit)?,
-  isScrollEnabled: Boolean,
```
L-Andrade, line 30: I've removed this from the arguments since it's already in `scrollState` which is also passed in another argument

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -106,12 +107,15 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
    * programmatic scrolling.
    */
   public var scrollHandler: ScrollHandler by
-    invalidatingObservable(ScrollHandler(themeHandler.scrollEnabled)) { oldValue, newValue ->
+    invalidatingObservable(
+      ScrollHandler(themeHandler.scrollEnabled, themeHandler.consumeMoveEvents)
```
L-Andrade, line 111: I've followed the implementation of `scrollEnabled` for the new `consumeMoveEvents`, but I'm not sure if this is the best choice

L-Andrade (commented):
Hello @Gowsky 
In the last commit (7bdbe27f912ae54e7849b08945a7880864f0e2ae), I've added a custom property to `VicoScrollState`: `consumeMoveEvents`.

Let me know if I missed anything, or if the new logic does not meet the existing implementation!

I have a video of the behaviour, but please take a look yourself too, of course:

https://github.com/user-attachments/assets/52549e23-1c4c-486a-8741-b8052eeb4040

Gowsky (changes requested):
Thanks for the changes. We have some suggestions:
- `consumeMoveEvents` shouldn't be in `VicoScrollState` or `ScrollHandler`, since it doesn't relate to scrolling.  Move gestures are used not just for scrolling, but also for dragging the marker. Thus it should be a parameter of `CartesianChartHost`, and a property of `CartesianChartView`.
- `consumeMoveEvents` is not needed in `MeasuringContext`.
- These changes need to be applied in the `multiplatform` module as well.

L-Andrade:
> Thanks for the changes. We have some suggestions:
> 
> - `consumeMoveEvents` shouldn't be in `VicoScrollState` or `ScrollHandler`, since it doesn't relate to scrolling.  Move gestures are used not just for scrolling, but also for dragging the marker. Thus it should be a parameter of `CartesianChartHost`, and a property of `CartesianChartView`.

Oh, that's fair. I've adjusted this in: a939fd3656b671a24202bb891dea334816c67439

> -  `consumeMoveEvents` is not needed in `MeasuringContext`.

Yeah, my bad! The view stuff threw me off a bit. I also addressed this in a939fd3656b671a24202bb891dea334816c67439

Let me know if this looks better. I've tested it briefly in Compose and Compose multiplatform, and both seemed to work as expected. Here's a short demo:

https://github.com/user-attachments/assets/6f66ba9f-b441-48bd-aa2d-00bd0ef79a60

Thanks for the feedback, and sorry for the delay!

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -325,6 +325,7 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
       if (chart.layerBounds.isEmpty) return@withChartAndModel
 
       motionEventHandler.scrollEnabled = scrollHandler.scrollEnabled
+      motionEventHandler.consumeMoveEvents = themeHandler.consumeMoveEvents
```
Gowsky, line 328: It should also be possible to toggle consumption programmatically. Thus, instead of updating  `consumeMoveEvents` with `themeHandler.consumeMoveEvents` here, a `consumeMoveEvents` property should be added to `CartesianChartView`. It should delegate the setter and the getter to `motionEventHandler.consumeMoveEvents`. `MotionEventHandler` should have a new constructor parameter for `consumeMoveEvents` which would be set using the `themeHandler`.

Gowsky (changes requested):
There's just one thing left that needs to be changed. After that, we will make minor tweaks and merge it.

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
L-Andrade: Makes sense. I've adjusted the constructor and added the changes in this commit: a92786e4ecf216f43b49cd44184010a6f27426f1

Let me know if anything's missing!

Gowsky: Everything's good now, thanks! We'll make some minor tweaks soon and merge the PR.

v01dn3ph1l1m:
@Gowsky when can we expect the merge? has been a while since I have been following this thread. 
Awaiting a release with this feature :)

Thanks a bunch for this library. its the best out there for compose multiplatform

Gowsky (approved):

Gowsky:
@L-Andrade, thank you for your contribution.

Hello, @v01dn3ph1l1m. We appreciate your kind words. The PR has just been merged, thus I think this answers your question. Cheers!

v01dn3ph1l1m:
@Gowsky thank you. can we have a version release for this?

shervinkoushan:
> @Gowsky thank you. can we have a version release for this?

Second that, would be great with a release @Gowsky :)

Gowsky:
[Vico `2.2.0-alpha.1`](https://github.com/patrykandpatrick/vico/releases/tag/v2.2.0-alpha.1) includes this change.

PR #1070, opened 2025-03-28
Scatter plot demo

ployt0:


ployt0:
This demonstrates the scatter plot often asked about and increasingly described by @Gowsky [658](https://github.com/patrykandpatrick/vico/discussions/658)

It uses multiple y-values for a single x, which is a limitation with naive line charts attempting to be scatter plots.

Gowsky:
Hello, @ployt0. Thanks for your contribution. The "Basic" sample charts are meant to showcase baseline usage of built-in layers, with effectively no customization, to be used as starters. In contrast, this is a repurposed line chart with nontrivial customization. Thus, it should be moved to the second group of sample charts and given real-world data. A "Basic scatter chart" sample chart can be added once a dedicated scatter layer is added (as planned).

The `Point` class seems redundant: the data is packed into `Point`s when created and then immediately unpacked for `series`.

Please reformat the code with `ktfmt --google-style`, as per the contribution guidelines.

ployt0:
I felt `Point` was more efficient than the key pairs I see used in other examples.

Believe it or not, this was based on (very boring) real data that I munged to fit with the other examples to have a limited continuous range of x values. Will just a rename do? The data was so boring it hardly seemed worth referencing it with a massive title/legend.

Where are the contribution guidelines and `ktfmt`?

ployt0:
Thanks for the constructive criticism. Hope this is acceptable.

ployt0:
https://github.com/user-attachments/assets/163d7e48-03c3-45fc-b818-251977161855

Very quick demonstration, don't blink!

ployt0:
https://github.com/user-attachments/assets/63b453cf-de33-4038-9a14-65a4d02c1b50

Very quick demonstration, in sensible format.

Gowsky:
> I felt `Point` was more efficient than the key pairs I see used in other examples.

Could you clarify how it's more efficient? It:

- Introduces an extra class.
- Makes the series definition longer and more complex (`listOf(Point(a, b), ...)` vs. simply `mapOf(a to b, ...)`).
- Makes the `Transaction` slower (two iterations of the `Point` list instead of access to precomputed key and value lists).

ployt0:
- The class is a single line
- It is preferable to have dedicated fields in frequently used constructs
- Vico itself uses the a `point.entry` class in a similar manner, rolling my own was the closest I had to hand.
- Maps make sense when there is a strict 1:1 mapping, which might be why vico, pre-scatter-plot favours them
- Even with a 1:1 mapping maps make more sense for random access, not sequential, which given we are populating a parameter named "series", is appropriate
- Maps have the overhead of hashing before they even see a list, at scale this is undesirable
- Regarding the two iterations, that's probably a valid point, I'm not so familiar with vico as yourself. As a newcomer, I find assigning to x and y series separately more explicit and readable.
- All my data is in lists, (or arrays), because it is faster to process and there is no 1:1 mapping

Gowsky:
> The class is a single line

The length of the definition alone doesn't determine the solution's efficiency, and if it did, this argument would still fail, since the alternative is zero lines. Furthermore, if we're to focus on code length, we should consider the length of the `data` expression too. With `Point`, it's longer and has more nesting.

The issue isn't the length of the `Point` definition but rather the presence of an extra construct that people referencing the sample chart must make themselves familiar with. Clarity is key here, and we try to use familiar, built-in Kotlin types when suitable.

> It is preferable to have dedicated fields in frequently used constructs

It's preferable when it carries a benefit. `points.map { it.x }` and `points.map { it.y }` are more complex than `points.keys` and `points.values`.

> Vico itself uses the a `point.entry` class in a similar manner, rolling my own was the closest I had to hand.

The context is different. Internally, `CartesianLayer.Entry` has the benefit of expressing a single _xy_ pair as a single object that can be easily passed around and has descriptive property names. In the sample chart, we're not reading individual _xy_ pairs, but rather passing the entire _x_ series and the entire _y_ series separately to `series`.

> Maps make sense when there is a strict 1:1 mapping, which might be why vico, pre-scatter-plot favours them

There is such a strict mapping here. It's required by `series`, where the data is passed.

> Even with a 1:1 mapping maps make more sense for random access, not sequential, which given we are populating a parameter named "series", is appropriate

Any performance differences are neglibile at this scale. On the other hand, logically, a map is suitable, since we're **mapping** _x_ values to _y_ values. This is of greater importance.

> Maps have the overhead of hashing before they even see a list, at scale this is undesirable

At the scale we're dealing with here, however, it makes no real difference. Of course, the same can be said about my comment about two list iterations, but that was an argument against the `Point` solution being faster (specifically, a counter for any similarly small performance advantage that the `Point` solution may have), not an argument for the suggested solution being more performant. Realistically, no performance difference will be observed between the two approaches, so code clarity should be the focus.

> Regarding the two iterations, that's probably a valid point, I'm not so familiar with vico as yourself. As a newcomer, I find assigning to x and y series separately more explicit and readable.

This doesn't relate to internal calculations or separating the _x_ and _y_ series. The two iterations come from the two `map` calls in your code: `points.map { it.x }` and `points.map { it.y }`.

> All my data is in lists, (or arrays), because it is faster to process and there is no 1:1 mapping

This is irrelevant in the context of this specific sample chart.

---

Due to the reasons outlined above and a desire to keep the sample charts consistent, we won't be able to merge this unless the suggested approach is taken.

ployt0:
Done.

I wasn't aware there was a consistency to be followed. Given the choice I made a choice of input, which we can agree to disagree on.

Gowsky:
>I wasn't aware there was a consistency to be followed.

Consistency is important in any codebase, especially in a library and its samples. However, this isn’t just about consistency for its own sake — first and foremost, the suggested approach is preferred for the reasons mentioned earlier. Aligning with it ensures both clarity and uniformity.

For the data, what does it represent, and what is the source?

ployt0:
I was not aware of any consistency. I don't find your arguments particularly convincing for any other reasons. They are your preferences and I respect them. The one of mine you so sweepingly dismissed is the most important, IMO, do unto others:

>> All my data is in lists, (or arrays), because it is faster to process and there is no 1:1 mapping

> This is irrelevant in the context of this specific sample chart.

I gave my best, if you don't like, I agree to disagree.

The data is the sales of freehold terraced houses in CV34 5DF, by week, in 2021. The data comes from the Land Registry. There are probably terms that require a whacking great notice of acknowledgement, but as its such a small sample I would seek exemption under artistic license, or something. Not a lawyer.

Gowsky:
Please avoid escalating the discussion or making it personal. Each of your points was addressed. I described the final one as irrelevant because it has no clear applicability here, which I believed to he apparent, especially given what I’d written above. To elaborate:

> All my data is in lists, (or arrays),

This is anecdotal.

> because it is faster to process

As mentioned earlier in my comment, no measurable performance difference will be observed in this particular case.

> and there is no 1:1 mapping

As mentioned earlier in my comment, in this particular case, there **is** such a mapping.

The argument wasn’t “sweepingly dismissed” — it was considered and found irrelevant for the aforementioned reasons. You’re more than welcome to ask for clarification.

My comments have been neutral and purely technical. This isn’t simply about personal preference — it’s about making decisions in the best interest of the project. Quoting maxims on good behavior and framing this as you “giving your best” and me “not liking it” is inappropriate.

Let’s keep the discussion professional and focused on the code. We’ll continue the review shortly.

ployt0:
All that is by the by. Had I seen the consistency I would of course have abided. While I'm happy with the result, I struggled to comprehend it on first reading, so gave my sample the way I did. I saved time by not studying every example.

Gowsky:
>The data is the sales of freehold terraced houses in CV34 5DF, by week, in 2021. The data comes from the Land Registry. 

We find this data too specific in terms of what it represents. We prefer the data that is broader and relatively easy to digest for the average person.

>There are probably terms that require a whacking great notice of acknowledgement, but as its such a small sample I would seek exemption under artistic license, or something.

We always provide the attribution, as seen in the other sample charts (except the basic ones, which use random data).

Furthermore, we've discussed the proposed solution and we prefer not to showcase workarounds in the sample app such as adding two series that act as one to showcase multiple _y_-values per _x_-value. While a scatter chart built with `LineCartesianLayer` is fine, we want to fit into what’s supported: either one series with one y-value per x-value, or multiple such series with distinct styling (not faking a single one). This is sufficient for a lot of data sets.

We plan to introduce a full support for scatter charts soon that will allow us to showcase more complex data scatter charts, including with multiple y-values per x-value.

Gowsky:
I'm closing this due to inactivity.

PR #1143, opened 2025-08-28
Add column click

husseinyehya687:


husseinyehya687:
add bar column click

Gowsky:
Hello, @husseinyehya687. Thank you for your contribution. Unfortunately, we have to close the PR due to the following issues:

1. The feature should be added to all built-in `CartesianLayer`s, not just `ColumnCartesianLayer`.
2. It should be added across all modules, not just to compose. The same applies to the sample charts.
3. This is too rigid, with insufficient abstraction. Checks like `layer is com.patrykandpatrick.vico.core.cartesian.layer.ColumnCartesianLayer` shouldn’t happen inside `CartesianChartHost`. This approach makes it impossible for custom `CartesianLayer`s to integrate with tap detection.

PR #1258, opened 2025-12-22
Synchronously restore cached `CartesianChartModel`

### LINKED ISSUE #1239: Chart flickers when starting a shared element transition
### Example

This zip contains a minimal project with a single Navigation 3 `NavDisplay` containing two screens that display the same chart using the same `ModelProducer` (with `animateIn` set to `false`). Clicking the button at the bottom of the screen navigates to the next screen, and swiping back navigates to the previous screen. The flicker can be easily seen while navigating.

[vico-shared-bounds-flicker-example.zip](https://github.com/user-attachments/files/24140131/vico-shared-bounds-flicker-example.zip)

### How to reproduce

Use `Modifier.sharedBounds(...)` on a chart inside a Navigation 3 `NavDisplay` (I haven't tested this with Navigation 2, but I think the behaviour should be the same) in 2 screens, set the `animateIn` parameter of each chart to false, and navigate between them.

### Observed behavior

The chart flickers just when starting the transition. The behaviour remains fine for the rest of the transition.

### Expected behavior

The chart does not flicker and animates smoothly.

### Module(s)

compose

### Platform(s)

Android

### Platform version(s)

Android 16 QPR2 (API 36.1)

### Vico version(s)

2.3.6, 2.4.0-alpha.3

### Additional information

Here's a screen recording of what a real implementation of this library in an app looks like with this bug, if it helps:

https://github.com/user-attachments/assets/545188d2-0b62-401a-9ce1-7bf1229a1c1b

BTW thank you so much for making this awesome library! I've been using it in my app and I've gotten really good feedback about the design of the app. I was able to style it so easily because of this library. Thanks again!

Gowsky:
This resolves #1239.

### DIFF
--- a/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianLayerModel.kt
+++ b/vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianLayerModel.kt
@@ -26,7 +26,8 @@ import kotlin.math.abs
 public interface CartesianLayerModel {
   /** Identifies this [CartesianLayerModel]. */
   @Deprecated(
-    "No longer used. If overriding this, remove the override, and implement equals and hashCode instead."
+    "No longer used. If overriding this, remove the override, and implement `equals` and " +
+      "`hashCode` instead."
   )
   public val id: Int
     get() = 0


vico/multiplatform/src/commonMain/kotlin/com/patrykandpatrick/vico/multiplatform/cartesian/data/CartesianChartModel.kt
```diff
@@ -138,6 +138,22 @@ internal fun CartesianChartModelProducer.collectAsState(
...
+    } else {
+      CartesianChartRanges.Empty
+    }
+  }
+
+  remember {
+    getCachedData(::updateRanges, extraStore)?.let { (model, ranges, extraStore) ->
+      dataState.set(model, ranges, extraStore)
+    }
+  }
```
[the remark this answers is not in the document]
Gowsky: We don't need to run this block when the chart changes. It's only relevant for state restoration and to avoid a flicker described in #1239.

vico/compose/src/main/java/com/patrykandpatrick/vico/compose/cartesian/CartesianChartModelProducer.kt
```diff
@@ -71,6 +74,22 @@ internal fun CartesianChartModelProducer.collectAsState(
...
+    } else {
+      CartesianChartRanges.Empty
+    }
+  }
+
+  remember {
+    getCachedData(::updateRanges, extraStore)?.let { (model, ranges, extraStore) ->
+      dataState.set(model, ranges, extraStore)
+    }
+  }
```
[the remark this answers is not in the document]
Gowsky: We don't need to run this block when the chart changes. It's only relevant for state restoration and to avoid a flicker described in #1239.

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianChartModel.kt
```diff
@@ -81,6 +81,12 @@ public class CartesianChartModel {
   public fun copy(extraStore: ExtraStore): CartesianChartModel =
     CartesianChartModel(models.map { it.copy(extraStore) }, id, width, extraStore)
 
+  override fun equals(other: Any?): Boolean =
+    this === other ||
+      other is CartesianChartModel && models == other.models && extraStore == other.extraStore
+
+  override fun hashCode(): Int = 31 * models.hashCode() + extraStore.hashCode()
```
patrickmichalik, line 89:
```suggestion
  override fun equals(other: Any?) =
    this === other ||
      other is CartesianChartModel && models == other.models && extraStore == other.extraStore

  override fun hashCode() = 31 * models.hashCode() + extraStore.hashCode()
```

vico/multiplatform/src/commonMain/kotlin/com/patrykandpatrick/vico/multiplatform/cartesian/data/CartesianChartModel.kt
```diff
@@ -110,6 +110,12 @@ public class CartesianChartModel {
   public fun copy(extraStore: ExtraStore): CartesianChartModel =
     CartesianChartModel(models.map { it.copy(extraStore) }, id, width, extraStore)
 
+  override fun equals(other: Any?): Boolean =
+    this === other ||
+      other is CartesianChartModel && models == other.models && extraStore == other.extraStore
+
+  override fun hashCode(): Int = 31 * models.hashCode() + extraStore.hashCode()
```
patrickmichalik, line 107:
```suggestion
  override fun equals(other: Any?) =
    this === other ||
      other is CartesianChartModel && models == other.models && extraStore == other.extraStore

  override fun hashCode() = 31 * models.hashCode() + extraStore.hashCode()
```

vico/multiplatform/src/commonMain/kotlin/com/patrykandpatrick/vico/multiplatform/cartesian/data/CartesianChartModelProducer.kt
```diff
@@ -127,6 +124,34 @@ public class CartesianChartModelProducer {
     updateReceivers.remove(key)
   }
 
+  internal fun getCachedData(
+    updateRanges: (CartesianChartModel?) -> CartesianChartRanges,
+    hostExtraStore: ExtraStore,
+  ): CachedData? =
```
patrickmichalik:
```suggestion
  ) =
```

vico/views/src/main/java/com/patrykandpatrick/vico/views/cartesian/CartesianChartView.kt
```diff
@@ -214,6 +208,24 @@ constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0
     }
   }
 
+  private fun restoreCachedModelData(): CartesianChartModel? =
```
patrickmichalik:
```suggestion
  private fun restoreCachedModelData() =
```

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianChartModel.kt
patrickmichalik, line 89: Do we certainly need both these and `CartesianChartModel.id`? The new functions consider `CartesianChartModel.extraStore`, while the `CartesianChartModel.id` calculation does not; additionally, `CartesianChartModel.copy` transfers the original `CartesianChartModel`’s ID to the new one. However, as far as I can tell, these two `CartesianChartModel.id` specifics aren’t functionally required. If anything, they’re tied to potentially unexpected behavior: `AutoScrollCondition.shouldScroll`, being called only when `CartesianChartModel.id` changes, is not invoked when a `CartesianChartModelProducer.Transaction` updates only the extras.

We should be able to deprecate `CartesianChartModel.id` in favor of the new functions. For consistency, we can also move away from `CartesianLayerModel.id` in favor of `equals` and `hashCode`, which the built-in `CartesianLayerModel`s already implement.

vico/multiplatform/src/commonMain/kotlin/com/patrykandpatrick/vico/multiplatform/cartesian/data/CartesianChartModel.kt
patrickmichalik, line 107: As [above](https://github.com/patrykandpatrick/vico/pull/1258#discussion_r2653536105), but breaking changes are acceptable in `multiplatform`.

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianChartModel.kt
Gowsky: The `id` property is now deprecated.

vico/multiplatform/src/commonMain/kotlin/com/patrykandpatrick/vico/multiplatform/cartesian/data/CartesianChartModel.kt
Gowsky: The `id` property is now removed.

vico/core/src/main/java/com/patrykandpatrick/vico/core/cartesian/data/CartesianLayerModel.kt
```diff
@@ -25,7 +25,11 @@ import kotlin.math.abs
 /** Stores a [CartesianLayer]’s data. */
 public interface CartesianLayerModel {
   /** Identifies this [CartesianLayerModel]. */
+  @Deprecated(
+    "No longer used. If overriding this, remove the override, and implement equals and hashCode instead."
```
patrickmichalik:
```suggestion
    "No longer used. If overriding this, remove the override, and implement `equals` and " +
      "`hashCode` instead."
```

PR #1371, opened 2026-02-28
Add Glance widget support for Cartesian charts

rubixhacker:
## Summary
This PR adds support for rendering Cartesian charts in Glance app widgets by introducing off-screen rendering capabilities and a new `compose-glance` module.

## Key Changes
- **New `renderToImageBitmap()` extension function** in `CartesianChartRenderer.kt`: Enables off-screen rendering of Cartesian charts to `ImageBitmap` objects. This function handles the complete rendering pipeline including range calculation, measurement, and drawing without requiring a Compose Canvas.

- **New `compose-glance` module**: A dedicated module for Glance widget integration that provides:
  - `CartesianChartImage()` composable: A Glance-compatible composable that renders charts as static images
  - Helper function to convert Compose `ImageBitmap` to Android `Bitmap` for Glance compatibility
  - Proper handling of density, layout direction, and font resolution for widget contexts

- **Module integration**: Updated build configuration and settings to include the new `compose-glance` module in the project structure and documentation generation.

## Implementation Details
- The off-screen rendering approach bypasses Glance's Canvas limitations by pre-rendering charts to bitmaps, making them suitable for widget display
- The implementation properly manages resources including cache cleanup via `CacheStore.purge()`
- Supports RTL layout direction detection from Android configuration
- Handles density and font scaling appropriately for different device configurations

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -18,6 +18,7 @@ lifecycleRuntime = "2.10.0"
 material = "1.13.0"
 mavenPublish = "0.36.0"
 mockK = "1.14.9"
+uiTextAndroid = "1.10.6"
 
 [libraries]
 activityCompose = { group = "androidx.activity", name = "activity-compose", version.ref = "activity" }
@@ -54,3 +55,4 @@ material = { group = "com.google.android.material", name = "material", version.r
 mavenPublish = { group = "com.vanniktech.maven.publish", name = "com.vanniktech.maven.publish.gradle.plugin", version.ref = "mavenPublish" }
 mockK = { group = "io.mockk", name = "mockk", version.ref = "mockK" }
 serialization = { group = "org.jetbrains.kotlin.plugin.serialization", name = "org.jetbrains.kotlin.plugin.serialization.gradle.plugin", version.ref = "kotlin" }
+uiTextAndroid = { group = "androidx.compose.ui", name = "ui-text-android", version.ref = "uiTextAndroid" }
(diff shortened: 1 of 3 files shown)


Gowsky:
Hello, @rubixhacker. Thanks for the PR. We've identified two issues:
- There's a compile error caused by the recent changes on the master branch. The error is caused by the missing `markerSeriesIndex` argument on `MutableCartesianMeasuringContext` instantiation.
- The new module `compose-glance` depends on both `views` and `compose` modules. They shouldn't be put together as users will only need one of them. However, we don't think that users using Glance will choose Vico `views` rather than Vico `compose` in their apps, thus the support for `views` can be dropped. We can re-add the support in a separate module if there's an interest in it.

vico/views/src/main/kotlin/com/patrykandpatrick/vico/views/cartesian/CartesianChartRenderer.kt
```diff
@@ -0,0 +1,96 @@
...
+ * This is intended for off-screen rendering (e.g., for use in app widgets).
+ *
+ * @param context the [Context] for density and font metrics.
+ * @param model the [CartesianChartModel] providing the chart data.
+ * @param width the width of the output [Bitmap] in pixels.
+ * @param height the height of the output [Bitmap] in pixels.
+ * @suppress
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public fun CartesianChart.renderToBitmap(
```
Gowsky: One last thing. This is no longer used and thus should be removed.

Gowsky (changes requested):

Gowsky:
The build fails due to a missing dependency. Adding `androidx.compose.ui:ui-text-android` to the `compose-glance` module should resolve it.

PR #1372, opened 2026-02-28
Add snap scrolling feature to align scroll position to grid windows

rubixhacker:
## Summary

This PR adds snap scrolling, which aligns the scroll position to multiples of a configurable _x_-axis step after the user stops scrolling. This is useful for charts where the viewport should settle on meaningful data boundaries. The feature is opt-in and applies to both stacks: Compose Multiplatform (all platforms) and Android Views.

## Public API

- **Compose:** `rememberVicoScrollState` and `VicoScrollState` gain an `xSnapStep: Double?` parameter (the snap step in _x_-axis units; `null` disables snapping) and a `snapAnimationSpec: AnimationSpec<Float>` parameter for the snap animation.
- **Views:** `ScrollHandler` gains an `xSnapStep: Double?` property plus `snapInterpolator` and `snapDuration` parameters. `snapDuration` is the animation duration when no fling velocity is available, and the maximum duration when a fling velocity is available.

## Snap calculation

Snap targets are computed by shared, side-effect-free helpers in the new `SnapScroll.kt` files (one per stack), keeping the Compose and Views logic in sync:

- `getSnapTarget(...)` converts `xSnapStep` to a pixel window using `xStep` and `xSpacing`, accounts for the layer's start padding so the grid aligns to data positions, and returns the nearest of the start (`0`), the closest snap step, and `maxValue` — all coerced into the valid scroll range.
- `getSnapDelta(...)` returns the delta from the current value to the snap target, or `null` when there's nothing to snap to (e.g. already aligned, or snapping disabled).

Both helpers accept a projected `targetValue` so snapping can target a fling's destination rather than the scroll position at release.

## Compose implementation

- A `SnapFlingBehavior` wraps the fling: it projects the decay's target value, then animates to the snap delta using `snapAnimationSpec`. It's installed via `Modifier.pointerInput` only when `xSnapStep` is set; otherwise the default fling behavior is used.
- `VicoScrollState` exposes internal `getSnapDelta()` and a `performSnap()` suspend function that stops any in-flight scroll, waits for it to settle, and animates to the snap target within a single mutator session.
- On Desktop and Web, `Modifier.extraPointerInput` calls `performSnap()` with the decay's projected target after a drag-driven fling instead of running the plain decay loop.

## Views implementation

- `MotionEventHandler` gains `onUserScroll`/`onSnapScroll` callbacks, fires `onUserScroll` only when a drag actually moves the scroll (so taps and sub-threshold movement don't trigger a snap), and, on fling, replaces the `OverScroller` fling with a `startScroll` to the snap target when one exists — using a velocity-derived duration from `ScrollHandler.getSnapScrollDuration`. It also stops any running snap animation on `ACTION_DOWN`.
- `ScrollHandler` gains `performSnap()` (used when no fling velocity is available), `getSnapTarget`, `getSnapScrollDuration`, and `stopAnimation`.
- `CartesianChartView` tracks user-scroll state (`wasUserScrolling`/`isUserScrolling`) and, in the animation loop, triggers `performSnap()` once the user has stopped scrolling and the scroller is no longer running.

## Tests

- `VicoScrollStateTest` (Compose) and `ScrollHandlerTest` (Views) cover start-padding alignment, the no-op-at-start case, and projected-target snapping.

### DIFF
--- a/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
+++ b/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
@@ -122,8 +122,8 @@ public class VicoScrollState {
    * @param autoScroll represents the scroll value or delta for automatic scrolling.
    * @param autoScrollCondition defines when an automatic scroll should occur.
    * @param autoScrollAnimationSpec the [AnimationSpec] for automatic scrolling.
-   * @param snapScrollX if not null, the scroll will snap to multiples of this _x_-axis window
-   *   width after the user stops scrolling.
+   * @param snapScrollX if not null, the scroll will snap to multiples of this _x_-axis window width
+   *   after the user stops scrolling.
    */
   public constructor(
     scrollEnabled: Boolean,
@@ -222,11 +222,13 @@ public class VicoScrollState {
     val snapScrollX = snapScrollX ?: return null
     val context = this.context ?: return null
     val layerDimensions = this.layerDimensions ?: return null
-    val windowPx =
-      (snapScrollX / context.ranges.xStep).toFloat() * layerDimensions.xSpacing
+    val windowPx = (snapScrollX / context.ranges.xStep).toFloat() * layerDimensions.xSpacing
     if (windowPx <= 0f) return null
+    val startPadding = layerDimensions.startPadding
     val snapTarget =
-      ((value / windowPx).roundToInt() * windowPx).coerceIn(0f.rangeWith(maxValue))
+      (startPadding + ((value - startPadding) / windowPx).roundToInt() * windowPx).coerceIn(
+        0f.rangeWith(maxValue)
+      )
     val delta = snapTarget - value
     return if (delta == 0f) null else delta
   }
--- a/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
+++ b/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
@@ -37,6 +37,7 @@ import androidx.compose.ui.geometry.Rect
 import com.patrykandpatrick.vico.compose.cartesian.data.CartesianChartModel
 import com.patrykandpatrick.vico.compose.cartesian.layer.CartesianLayerDimensions
 import com.patrykandpatrick.vico.compose.common.rangeWith
+import kotlin.math.roundToInt
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.first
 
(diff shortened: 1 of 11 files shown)


vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
```diff
@@ -193,30 +194,39 @@ public class VicoScrollState {
...
+        animationSpec,
+      )
     }
   }
 
+  internal fun getSnapDelta(): Float? {
+    val snapScrollX = snapScrollX ?: return null
+    val context = this.context ?: return null
+    val layerDimensions = this.layerDimensions ?: return null
+    val windowPx =
```
Gowsky: This should also account for `CartesianLayerDimensions.scalableStartPadding` and `CartesianLayerDimensions.unscalableStartPadding`.

vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
```diff
@@ -118,19 +115,23 @@ public class VicoScrollState {
...
+   * @param snapScrollX if not null, the scroll will snap to multiples of this _x_-axis window
+   *   width after the user stops scrolling.
    */
   public constructor(
     scrollEnabled: Boolean,
     initialScroll: Scroll.Absolute,
     autoScroll: Scroll,
     autoScrollCondition: AutoScrollCondition,
     autoScrollAnimationSpec: AnimationSpec<Float>,
+    snapScrollX: Double? = null,
```
Gowsky: The name should be more meaningful, but we can handle it at a later stage after merging it.

Gowsky (changes requested):
Hello, @rubixhacker. Thank you for your contribution. The PR looks good overall, but please address the review comments.

vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
rubixhacker: Done, snap target is now startPadding + round((value - startPadding) / windowPx) * windowPx, coerced into [0, maxValue], so the snap grid aligns to data positions instead of the raw scroll origin. Same change applied to the Views ScrollHandler.performSnap for parity

vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollState.kt
rubixhacker: How about xSnapInterval?

Gowsky: Let's go with `xSnapStep`.

patrickmichalik: Done in [`ad9400f`](https://github.com/patrykandpatrick/vico/pull/1372/commits/ad9400fdd184e3578a89a55ed8b41059c0c33904).

patrickmichalik (approved):
Thanks!

PR #1419, opened 2026-04-11
Improve `PointerInputChange` consumption logic

Gowsky:
`androidx.compose.foundation.gestures.detectTapGestures` used for detecting tap and long press interactions has caused an excessive `PointerInputChange` consumption leading to other `detectTapGestures` higher up the `Modifier` chain to be non functional. 

This resolves #1391.

## Interactions whose handling changes

The `onInteraction` callback signature changes from `(Interaction) -> Unit` to `(Interaction) -> Boolean`. The return value indicates whether the event was handled and its `PointerInputChange` should be consumed. In `CartesianChartHostImpl`, this returns `true` when `MarkerController.shouldAcceptInteraction` accepts the interaction, and `false` otherwise.

- **Tap** — Previously detected via `androidx.compose.foundation.gestures.detectTapGestures`, which unconditionally consumed pointer input. Now uses a custom implementation (`awaitFirstDown` + `waitForUpOrCancellation`) that only consumes the `PointerInputChange` when `onInteraction` returns `true`.
- **LongPress** — Same as Tap: replaces `detectTapGestures` with a custom implementation (`awaitLongPressOrCancellation`) that conditionally consumes based on the `onInteraction` return value.
- **Press** — Never consumed (`false` from the consumption check).
- **Release** — Never consumed (`false` from the consumption check).
- **Move** — Consumed only when `consumeMoveEvents && !scrollState.scrollEnabled` (previously consumed via `PointerInputChange.consume()` under the same condition, but inline rather than via the unified `consume` flag).
- **Enter** — Consumed based on the `onInteraction` return value.
- **Exit** — Consumed based on the `onInteraction` return value.
- **Scroll (Zoom)** — Always consumes all pointer changes when zoom fires.

### DIFF
--- a/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/Modifier.kt
+++ b/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/Modifier.kt
@@ -31,6 +31,7 @@ import androidx.compose.ui.input.pointer.PointerInputChange
 import androidx.compose.ui.input.pointer.PointerInputScope
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.unit.IntSize
+import androidx.compose.ui.util.fastForEach
 import com.patrykandpatrick.vico.compose.cartesian.marker.Interaction
 import com.patrykandpatrick.vico.compose.common.Point
 import com.patrykandpatrick.vico.compose.common.detectZoomGestures
@@ -46,7 +47,7 @@ private fun Offset.toPoint() = Point(x, y)
 @Composable
 internal fun Modifier.pointerInput(
   scrollState: VicoScrollState,
-  onInteraction: ((Interaction) -> Unit)?,
+  onInteraction: ((Interaction) -> Boolean)?,
   onZoom: ((Float, Offset) -> Unit)?,
   consumeMoveEvents: Boolean,
   longPressEnabled: Boolean,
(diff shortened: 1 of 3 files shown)


vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/Modifier.kt
```diff
@@ -111,4 +133,43 @@ internal fun Modifier.pointerInput(
...
+) {
+  awaitEachGesture {
+    val down = awaitFirstDown()
+    val inputChange =
+      if (onLongPress != null) {
+        val longPress = awaitLongPressOrCancellation(down.id)
+        if (longPress != null) {
+          if (onLongPress(longPress.position)) {
+            longPress.consume()
+          }
```
patrickmichalik:
```suggestion
          if (onLongPress(longPress.position)) longPress.consume()
```

vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/Modifier.kt
```diff
@@ -111,4 +133,43 @@ internal fun Modifier.pointerInput(
...
+      }
+    }
+  }
+}
+
+@OptIn(ExperimentalContracts::class)
+context(pointerEventScope: AwaitPointerEventScope)
+private fun PointerInputChange?.isTap(firstDown: PointerInputChange): Boolean {
+  contract { returns(true).implies(this@isTap != null) }
+  this ?: return false
```
patrickmichalik:
```suggestion
  if (this == null) return false
```

vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/Modifier.kt
```diff
@@ -111,4 +133,43 @@ internal fun Modifier.pointerInput(
...
+        }
+        currentEvent.changes.firstOrNull { it.id == down.id }
+      } else {
+        waitForUpOrCancellation()
+      }
+    if (inputChange.isTap(down)) {
+      if (onTap(inputChange.position)) {
+        inputChange.consume()
+      }
+    }
```
patrickmichalik:
```suggestion
    if (inputChange.isTap(down) && onTap(inputChange.position)) inputChange.consume()
```

patrickmichalik (approved):

Issue #1426, opened 2026-04-16
Charts with Absolute.End Autoscroll Flash First Page When Previous Data Is Added


bcapps:
### Example

The example app linked here in the `vico-backfill-repro` folder shows the type of chart that might run into this issue: in a chart of money over time, for example, we might minimally have one viewport page of data that doesn't scroll, and choose to load a longer history of more leading pages, but stay scrolled to the trailing edge when that data loads in for a seamless experience. However, you'll see that it quickly flashes to the oldest page of data with a line of 0 y value.

https://github.com/bcapps/vico/tree/repro/backfill-transition-sample/repros/vico-backfill-repro

### How to reproduce

In the sample app, screen record a video and wait several seconds. You'll see the flash as it transitions from one page of data to 4 pages.

### Observed behavior

You may see the flash, but the video will make clear that this flash is actually one frame of the beginning/first page of the chart with a line showing all 0 y values before flashing back to the last page.

### Expected behavior

No flash occurs and the chart stays scrolled to the last page without flashing to the first page.

### Module(s)

compose

### Platform(s)

Android

### Platform version(s)

16.1

### Vico version(s)

3.1.0

### Additional information

`CartesianChartHost` currently schedules `scrollState.autoScroll(...)` in a coroutine after updating the new model and layout state.

For `Scroll.Absolute.End` with a snap animation, that allows one frame where the new model can render at the old scroll offset before the snap is applied. In cases where we start with some data and load previous pages of history, this can briefly expose the wrong (first page) viewport for a single frame.

patrickmichalik:
Hello! Thanks for the report. We’re looking into it.

bcapps:
Thanks very much for taking a look and addressing so quickly!

bcapps:
@patrickmichalik Thanks again for the fix, just wondering they the versioning went from 3.1.0 down to 2.4.4 for the latest release?

PR #1427, opened 2026-04-16
Apply snapped auto-scroll before first draw on model growth

bcapps:
## Summary

This PR proposes one possible fix for #1426.

In Compose `CartesianChartHost`, when a chart model grows and auto-scroll is configured to snap to the end, the new model can currently render once at the previous scroll offset before the auto-scroll coroutine runs. In single-page -> backfilled-history transitions, that can produce a visible one-frame jump.

This change adds a synchronous fast path for snapped auto-scroll:
- if `autoScrollAnimationSpec` is a `SnapSpec`
- and `autoScrollCondition` says the new model should scroll

then the scroll delta is applied immediately before the first draw of the new model.

For non-snap cases, the existing asynchronous path is preserved.

## Why this change

Today, `CartesianChartHost` updates the model and layout state first, then launches `scrollState.autoScroll(...)` in a coroutine. That separation is reasonable for animated scrolling, but for snapped scrolling it leaves a stale-frame window where the chart can briefly draw at the wrong viewport.

This PR narrows the fix to the snapped case only. It keeps the current behavior for animated auto-scroll, while avoiding the stale first frame for `Scroll.Absolute.End`-style updates.

## Scope

- Compose module only
- no public API changes
- no change to animated auto-scroll behavior
- fixes one specific timing issue around snapped auto-scroll

## Testing

- verified against a standalone repro that simulates a cached trailing page being replaced by a wider historical model through the same `CartesianChartModelProducer`

Standalone repro:
https://github.com/bcapps/vico/tree/repro/backfill-transition-sample/repros/vico-backfill-repro

## Notes

This is just one possible fix, but let me know if it's too specific for this use case.

### DIFF
--- a/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
+++ b/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/CartesianChartHost.kt
@@ -253,7 +253,9 @@ internal fun CartesianChartHostImpl(
     scrollState.update(measuringContext.value, chart.layerBounds, layerDimensions)
 
     if (model != lastHandledModel) {
-      coroutineScope.launch { scrollState.autoScroll(model, previousModel) }
+      if (!scrollState.autoScrollBeforeFirstDraw(model, previousModel)) {
+        coroutineScope.launch { scrollState.autoScroll(model, previousModel) }
+      }
       lastHandledModel = model
     }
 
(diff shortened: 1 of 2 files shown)


Gowsky:
Hello, @bcapps. Thank you for your contribution. I've looked through the changes and tested them using the MRE provided under #1426. Due to some issues in the proposed fix, such as code duplication or incorrect assumption that `SnapSpec<T>` has no delay, I've opened #1429 with a bit more flexible approach.

bcapps:
> Hello, @bcapps. Thank you for your contribution. I've looked through the changes and tested them using the MRE provided under #1426. Due to some issues in the proposed fix, such as code duplication or incorrect assumption that `SnapSpec<T>` has no delay, I've opened #1429 with a bit more flexible approach.


Excellent, thank you so much!

PR #1482, opened 2026-05-15
Fix: use dataLabelValueFormatter in rememberLine()

jared-loseit:
dataLabelRotationDegrees is passed to remember twice and dataLabelValueFormatter is not passed to remember


patrickmichalik (approved):
Thanks!

PR #1484, opened 2026-05-16
feat: add edge-aligned item placer for horizontal axis labels

nagatsuka-shuuya:
Closes https://github.com/patrykandpatrick/vico/discussions/1475

### DIFF
--- a/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/axis/HorizontalAxis.kt
+++ b/vico/compose/src/commonMain/kotlin/com/patrykandpatrick/vico/compose/cartesian/axis/HorizontalAxis.kt
@@ -192,8 +192,22 @@ protected constructor(
               layoutDirectionMultiplier
         val previousX = labelValues.getOrNull(index - 1) ?: (fullXRange.start.doubled - x)
         val nextX = labelValues.getOrNull(index + 1) ?: (fullXRange.endInclusive.doubled - x)
+        val shiftExtremeLabels = itemPlacer.getShiftExtremeLabelsInward(this)
+        val horizontalPosition =
+          when {
+            shiftExtremeLabels && index == 0 -> Position.Horizontal.End
+            shiftExtremeLabels && index == labelValues.lastIndex -> Position.Horizontal.Start
+            else -> Position.Horizontal.Center
+          }
         val maxWidth =
-          ceil(min(x - previousX, nextX - x) / ranges.xStep * layerDimensions.xSpacing).toInt()
+          when (horizontalPosition) {
+            Position.Horizontal.End ->
+              ceil((nextX - x) / ranges.xStep * layerDimensions.xSpacing).toInt()
+            Position.Horizontal.Start ->
+              ceil((x - previousX) / ranges.xStep * layerDimensions.xSpacing).toInt()
+            else ->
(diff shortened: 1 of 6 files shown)


Gowsky:
Thanks for the contribution @akiya-nagatsuka, and for addressing the earlier review feedback — the rename to `extremes()`, the boolean flag, and the consistent `getLabelValues`/`getLineValues` are all the right calls. A few remaining things to address:

### 1. `getShiftExtremeLabelsInward` called inside the per-label loop

In both `HorizontalAxis.kt` files the call sits inside the label-drawing loop, so it's invoked once per label even though the result never changes between iterations. It should be hoisted above the loop:

```kotlin
val shiftExtremeLabels = itemPlacer.getShiftExtremeLabelsInward(this)
labelValues.forEachIndexed { index, x ->
    val previousX = ...
    val nextX     = ...
    val horizontalPosition = when {
        shiftExtremeLabels && index == 0                     -> Position.Horizontal.End
        shiftExtremeLabels && index == labelValues.lastIndex -> Position.Horizontal.Start
        else                                                 -> Position.Horizontal.Center
    }
    ...
}
```

### 2. `getWidthMeasurementLabelValues` returns values that are never consumed

`ExtremesHorizontalAxisItemPlacer.getWidthMeasurementLabelValues` returns `context.ranges.measuredLabelValues`, but the resulting `maxLabelWidth` flows nowhere: `getLabelValues`/`getLineValues` ignore it, `getFirstLabelValue`/`getLastLabelValue` return `null`, and `getStartLayerMargin`/`getEndLayerMargin` don't reference the `maxLabelWidth` parameter. The label formatting and measurement triggered by a non-empty return is wasted work at every layout pass.

This should return `emptyList()`. (If label-width measurement is ever needed here in the future, the right sample would be `listOf(ranges.minX, ranges.maxX)` — `measuredLabelValues` is not suitable because its second element is the last *step-aligned* value, which may not equal `maxX`.)

### 3. Dead subtraction in `getStartLayerMargin`/`getEndLayerMargin`

The margin overrides subtract `layerDimensions.unscalableStartPadding`/`unscalableEndPadding`, copying `AlignedHorizontalAxisItemPlacer`. But those values are set by `updateLayerDimensions` only when `getFirstLabelValue`/`getLastLabelValue` return non-null — and `ExtremesHorizontalAxisItemPlacer` always returns `null` for both. The subtraction always operates on 0. The bodies can be simplified to match `SegmentedHorizontalAxisItemPlacer`:

```kotlin
override fun getStartLayerMargin(...): Float =
    if (shiftExtremeLines) tickThickness else tickThickness.half
```

### 4. `aligned(shiftExtremeLabelsInward = true, addExtremeLabelPadding = true)` interaction undocumented

Using both flags together produces a confusing result: `addExtremeLabelPadding` sets `unscalableStartPadding`/`unscalableEndPadding` (via `getFirstLabelValue`/`getLastLabelValue`), which then reduces the start/end layer margins — while `shiftExtremeLabelsInward` makes that padding unnecessary in the first place. The KDoc on `aligned()` should note that `shiftExtremeLabelsInward = true` is intended as an *alternative* to `addExtremeLabelPadding = true`, and that combining both is not recommended.

### 5. Nit: missing blank line in `AlignedHorizontalAxisItemPlacer`

After inserting the `getShiftExtremeLabelsInward` override, `getSpacingOrThrow` follows immediately without a blank line. The rest of the file separates members consistently — please align.

patrickmichalik (approved):
Thanks!

PR #1488, opened 2026-05-24
Add animateZoom to Compose and Views stacks

nagatsuka-shuuya:
Implements #1485.

## Summary

- Adds `VicoZoomState.animateZoom` (Compose) and `ZoomHandler.animateZoom` (Views), mirroring the existing `animateScroll` API. Both accept a target `Zoom`, animation parameters (an `AnimationSpec` in Compose; a duration and `TimeInterpolator` in Views), and an optional `bias` — the zoom anchor’s horizontal position within the `CartesianChart` bounds, from 0 (the start edge) to 1 (the end edge), defaulting to 0.5. Passing 1 keeps the right edge fixed during the animation.

- Includes a `ZoomTransitionLineChart` sample demonstrating a sliding-window viewport transition: a superset dataset is loaded once, and period toggles (1W / 1M / 3M) animate the zoom with the right edge anchored to the latest data point. The Y axis scales dynamically to the visible period’s data range.

## Demo

[Screen_recording_20260607_003813.webm](https://github.com/user-attachments/assets/1c1f19f2-0ebe-481d-86ea-a554c29b9509)


Gowsky (dismissed):
I've verified the concern about phantom entries. They aren't persisted indefinitely, so everything's fine. Thank you for your contribution, @akiya-nagatsuka!

Gowsky:
To put more context to the `ItemPlacer`s issue, the key point is that the ranges shouldn't be interpolated. This causes a bit unexpected behavior and will interfere with some `ItemPlacer`s implementations. Instead, we'd recommend moving the positioning logic to the drawing models, just like it's already done for _y_-range changes.

nagatsuka-shuuya:
@Gowsky  Hi, I've reworked the implementation based on your feedback:

-  Removed range interpolation — axes and ItemPlacers now always receive the final range, which fixes the layout shift issue

- Moved x-positioning into the drawing model — each entry stores an x-fraction that lerps between old and new positions during animation

- New points are seeded with a start position in the previous range's coordinate space so they slide in from outside the chart with consistent easing

Could you take a look and let me know if this is the right direction?

dmitry-stakhov:
@claude 
> It'd help to know which scenario this is really for

Currently we are trying to replace the HighChart library in our project. We have the user assets plot with different time interval like 1 month, 3 months, 6 month, 1 year and all time. So the data is overlapping for these time periods. For example, the data for last 3 months is also included in data for last 6 months.

> But the sample in this PR is something else. recentData and historicalData are the same series shifted — the overlapping region is identical — i.e. a sliding window over a single dataset.

This is exactly our use case, we need to shift from `recentData` and `historicalData` to show different time periods.
Currently, in our project implementation we have 5 maps of X-Y data and when the user switches between time periods we pass related map of data.

> the camera (scroll + an animated zoom)

This should also fulfill our use-case, though we need some refactoring to moving from using 5 separate maps of data to using one the biggest map of data + adding zoom/camera movement.

I checked that the sample added in this PR implements our plot current appearance and we are looking forward for such animations being added.
I am good with both implementation alternatives if they help to achieve the same goal. Please let me know if any other details about the use-case are needed.

patrickmichalik:
I’m marking this as a draft as we pinpoint the requirements. Thanks for sticking with us; the above-requested screen recording should settle the matter.

nagatsuka-shuuya:
@patrickmichalik Sure, let me know if anything from my side is needed. Thank you for taking a look

patrickmichalik:
Thanks, @akiya-nagatsuka! I think your reply may have crossed with [the comment just above](https://github.com/patrykandpatrick/vico/pull/1488#issuecomment-4638234230). As described there, this is a viewport transition, and animated zoom looks like the actual missing primitive for your use case. Would you like to prototype that piece and try it against your period-switching setup?

nagatsuka-shuuya:
@patrickmichalik Thank you for pointing, for some reason claude message didn't load till now. Adding animated zoom sounds good to me, let me try add it. Can I also include the fix for Y-axis max value into the PR or should I keep it on my project side?

nagatsuka-shuuya:
@claude @patrickmichalik @Gowsky

Thanks for the feedback. I revised the implementation and added `animateZoom` to both stacks, along with a sample demonstrating the sliding-window use case.  I've also updated the PR description and attached a demo video.

While building the sample I noticed that `ExtremesHorizontalAxisItemPlacer` always shows `minX`/`maxX` of the full dataset rather than the visible viewport edges - so I implemented a custom item placer for the sample. Happy to fix `ExtremesHorizontalAxisItemPlacer` itself if that's the right direction.

Please take a look - I'm open to any follow-up fixes or improvements.

nagatsuka-shuuya:
@claude I created https://github.com/patrykandpatrick/vico/pull/1513 to fix `ExtremesHorizontalAxisItemPlacer` issue

patrickmichalik:
Cheers! I’ve just merged that.

nagatsuka-shuuya:
@patrickmichalik Thanks, I updated this branch to use the existing `ExtremesHorizontalAxisItemPlacer` and double checked that it works properly now.
The sample I added represents my use-case, but feel free to update to make it more general if it's needed.

patrickmichalik (approved):
Cheers! Made some tweaks in [`3711d64`](https://github.com/patrykandpatrick/vico/pull/1488/commits/3711d64ec935a1a0afc04ebf590e1ac65380eecb). Thanks for your patience.

PR #1489, opened 2026-05-26
Add configurable axis title positions

hjf-codecrazy:
Adds configurable title positions for horizontal and vertical axes.
ABC

### DIFF
--- a/buildSrc/src/main/kotlin/publishing-convention.gradle.kts
+++ b/buildSrc/src/main/kotlin/publishing-convention.gradle.kts
@@ -17,8 +17,8 @@
 plugins { id("com.vanniktech.maven.publish") }
 
 mavenPublishing {
-  //publishToMavenCentral(automaticRelease = true)
-  //signAllPublications()
+  publishToMavenCentral(automaticRelease = true)
+  signAllPublications()
   pom {
     name = "Vico"
     description = "A powerful and extensible multiplatform chart library."
--- a/buildSrc/src/main/kotlin/publishing-convention.gradle.kts
+++ b/buildSrc/src/main/kotlin/publishing-convention.gradle.kts
@@ -17,8 +17,8 @@
 plugins { id("com.vanniktech.maven.publish") }
 
 mavenPublishing {
-  publishToMavenCentral(automaticRelease = true)
-  signAllPublications()
+  //publishToMavenCentral(automaticRelease = true)
+  //signAllPublications()
   pom {
     name = "Vico"
     description = "A powerful and extensible multiplatform chart library."
--- a/guide/v3.x.x/compose/cartesian-charts/axis.md
+++ b/guide/v3.x.x/compose/cartesian-charts/axis.md
@@ -23,6 +23,8 @@ Numerous customization options are available; you can change the appearance of t
 
 [`BaseAxis.TickPosition`](https://api.vico.patrykandpatrick.com/vico/compose/com.patrykandpatrick.vico.compose.cartesian.axis/-base-axis/-tick-position/) and [`BaseAxis.LineDrawingOrder`](https://api.vico.patrykandpatrick.com/vico/compose/com.patrykandpatrick.vico.compose.cartesian.axis/-base-axis/-line-drawing-order/) let you place ticks outside, inside, or across the axis line and choose whether ticks and the axis line are drawn under or over the chart layers.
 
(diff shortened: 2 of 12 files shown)


hjf-codecrazy:
> Thanks for the contribution! This is a nice addition, and the dual-stack coverage (Compose + Views) plus the XML attribute wiring is appreciated. A few things stand out before this can be merged:
> 
> ### Must-fix: unrelated changes
> Two unrelated changes look like they came from a local development setup and should not land here:
> 
> * [`buildSrc/src/main/kotlin/publishing-convention.gradle.kts`](https://github.com/patrykandpatrick/vico/pull/1489/files#diff-09d6bdc4b9a2bc6df9c0e6a18ce20aa6ac11afdcc4c9b9d97d31d33fb89bd6ee) comments out `publishToMavenCentral(automaticRelease = true)` and `signAllPublications()`. This would break Maven Central publishing.
> * [`jitpack.yml`](https://github.com/patrykandpatrick/vico/pull/1489/files#diff-d6f8e96cfb50dd4ee18b3f7e58d8c4d2f4c45e7e8c5a26b86a08d0c0c0c2c0c0) is new — the library is published via Maven Central, not JitPack.
> 
> Please drop both files from this PR.
> 
> ### API consistency between `HorizontalAxis` and `VerticalAxis`
> Parameter placement for `titlePosition` differs between the two classes:
> 
> * In `HorizontalAxis`, `titlePosition` is inserted right after `itemPlacer` (in the primary constructor, the internal secondary constructor, `copy()`, `rememberTop`/`rememberBottom`, and `top()`/`bottom()`).
> * In `VerticalAxis`, it’s inserted near the end, after `title` and before `tickPosition`.
> 
> Please pick one position and apply it symmetrically across both classes. This affects the public API signatures (factory functions, `copy()`), so consistency matters for source clarity.
> 
> Related: in the internal secondary constructor, `HorizontalAxis` gives `titlePosition` a default of `TitlePosition.Center`, but `VerticalAxis` has no default. They should match — and given that secondary constructors generally don't carry defaults in this codebase, removing the default from `HorizontalAxis` is probably the right move.
> 
> ### `VerticalAxis.TitlePosition.Top` layout
> The `Top` title is drawn centered horizontally on `getAxisLineCenterX()` with `maxWidth = canvasSize.width.toInt()`. Since the axis line sits at the chart edge (`bounds.right - lineThickness.half` for a `Start` axis, mirror for `End`), a non-trivial title will overflow off the canvas on one side and into the chart area on the other. Two concerns:
> 
> 1. For typical titles this places half the text outside the canvas. Centering over the layer area (or anchoring with `Position.Horizontal.Start`/`End` aligned to the canvas edge) would render more usefully.
> 2. If both a `Start` and an `End` axis use `Top`, both titles draw at `y = bounds.top` and will collide. Worth either resolving this or documenting the limitation in the `TitlePosition.Top` KDoc.
> 
> ### `HorizontalAxis` draw order
> In both Compose and Views `HorizontalAxis.drawUnderLayers`, the title draw moved from inside the `canvas.save()`/`restore()` block to after the restore (i.e., outside the clip path). This is needed for `End` so the title isn’t clipped to the line range, but it’s a behavior change for the existing `Center` rendering — please verify that titles that previously relied on being clipped (e.g., very wide titles next to layer bounds) still render acceptably.
> 
> ### Minor
> * The new `getAxisLineCenterX()` helper duplicates the inline `lineThickness.half` math already in `drawLineAndTicks` (VerticalAxis.kt L272–L274 / equivalent in Compose). Not a blocker, but a single helper used in both places would be tidier.
> * KDoc for `TitlePosition.End` says “centered vertically on the line.” The implementation uses `bounds.bottom - lineThickness.half` for top axes and `bounds.top + lineThickness.half` for bottom axes — i.e., centered on the axis line, which matches; just worth double-checking the wording reads as intended.
> * Consider adding a brief guide-page mention of `titlePosition` so it’s discoverable alongside the existing axis customization docs.
> 
> Happy to take another pass once the unrelated files are dropped and the constructor signatures line up.

Thanks for the detailed review! I've addressed the must-fix items:
✅ Removed unrelated files: Deleted jitpack.yml and buildSrc/src/main/kotlin/publishing-convention.gradle.kts in commit "Remove unrelated files from PR"
The remaining items (API consistency between HorizontalAxis and VerticalAxis, TitlePosition.Top layout, draw order, and minor improvements) are being worked on. I'll push those fixes shortly and update this PR.

dmitry-stakhov:
Closes https://github.com/patrykandpatrick/vico/discussions/1478

patrickmichalik (approved):
Thanks!

Issue #1369, opened 2026-02-27
ScrollState after updating candlestick graph


antonis-bc:
### Example

I have a candlestick graph initialised with a scrollState
```
 val scrollState = rememberVicoScrollState(
        initialScroll = Scroll.Absolute.End,
    )
```
Initially i am receiving from our API a set of candles which are getting rendered normally on my graph.

What i want to achieve is when the user scrolls the graph to the older rendered candle request more older candles from our api, pre-append to the existing ones. So what i do i to observe the scrollState value to get to 0.
This is all work and older data are coming from our API. 
The that i have is that when these data are getting rendered on the graph, the scrollstate that is 0, makes the graph to scroll to the new oldest candle, causing another request to our API for older data.

### How to reproduce

Create a candlestick graph with e.g 200 candles with initialScroll set to End and when this reaches 0 add another batch of 200 older candles 

### Observed behavior

The graph scrolls automatically to the updated oldest candle

### Expected behavior

Not sure if this the expected behavior but what i want to achieve is append the data on the graph, without any additional scrolling when the new data are getting append on the graph. Just keep the graph visible state as it was before the "earlier" candles append

### Module(s)

compose, compose-m3

### Platform(s)

Android

### Platform version(s)

Andorid 16

### Vico version(s)

3.0.1

### Additional information

https://github.com/user-attachments/assets/4c5e55a6-b2fb-4390-ad71-8c3d8db1c618

PR #1539, opened 2026-06-29
Preserve the visible x range across data updates

lewan42:
**Problem.** When data is added on the left (e.g., older history is prepended), the chart jumps: the viewport shifts to the new edge instead of staying on the data that was on screen. For infinite scroll this is a dealbreaker — jumping to the start edge immediately triggers another load, so paging in older history is unusable. There's currently no built-in way to keep the visible range stable across data updates.

**Solution.** `VicoScrollState` now preserves the visible x range across data updates. When the chart's x range shifts (its `minX` or `xStep` changes), the scroll value is recomputed so the same x coordinates stay on screen and the start edge of the visible range stays anchored to the same data. The correction happens in `VicoScrollState.update` during measurement, so the adjusted value is used by the same frame that draws the new data — no flicker, no post-layout scroll. Auto-scrolling still applies afterward, and the behavior works in both LTR and RTL layouts.

Addresses the use case in #1369.

**Before / after**
| Before | After |
| :--- | :--- |
| https://github.com/user-attachments/assets/a8a7cb69-0244-4431-b07c-1a4239002edc | https://github.com/user-attachments/assets/cbb3b5d5-ff52-47d9-aeeb-5f236c5658fa |


lewan42:
Thanks for spotting the cleaner approach — totally agree it's more of a bug than a feature worth keeping. Reworked everything to match: dropped the enum and the    dataUpdateScrollAnchor param, and made the visible-x-range correction the default behavior. Was glad to contribute to the library! 🙂

patrickmichalik (approved):
Thanks!

PR #1547, opened 2026-07-10
feat(PieChart): add rotation via drag

adrcotfas:
The commit message is self explanatory. Here's a demo:

[Screen_recording_20260710_141829.webm](https://github.com/user-attachments/assets/bc5d9ad9-73db-49d8-890a-c50bb51be165)

---
<a href="https://app.blacksmith.sh/patrykandpatrick/codesmith/vico/pr/1547"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786274422&installation_id=136960981&pr_number=1547&repository=patrykandpatrick%2Fvico&return_to=https%3A%2F%2Fgithub.com%2Fpatrykandpatrick%2Fvico%2Fpull%2F1547&signature=9f21486d99de5dc240692780c18ddb9ae4bc3b12aaedf9cfc7545b2ba8e57786"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/sample/shared/src/commonMain/kotlin/com/patrykandpatrick/vico/sample/charts/BasicPieChart.kt
+++ b/sample/shared/src/commonMain/kotlin/com/patrykandpatrick/vico/sample/charts/BasicPieChart.kt
@@ -31,13 +31,16 @@ import com.patrykandpatrick.vico.compose.pie.PieChartHost
 import com.patrykandpatrick.vico.compose.pie.data.PieChartModelProducer
 import com.patrykandpatrick.vico.compose.pie.data.PieValueFormatter
 import com.patrykandpatrick.vico.compose.pie.data.pieModel
+import com.patrykandpatrick.vico.compose.pie.pieRotation
 import com.patrykandpatrick.vico.compose.pie.rememberPieChart
+import com.patrykandpatrick.vico.compose.pie.rememberPieRotationState
 
 @Composable
 private fun ComposeBasicPieChart(
   modelProducer: PieChartModelProducer,
   modifier: Modifier = Modifier,
 ) {
+  val rotation = rememberPieRotationState()
   PieChartHost(
     chart =
       rememberPieChart(
@@ -53,10 +56,11 @@ private fun ComposeBasicPieChart(
               )
             }
           ),
+        startAngle = -90f + rotation.angle,
         valueFormatter = PieValueFormatter { _, value, _ -> "${value.toInt()}%" },
       ),
     modelProducer = modelProducer,
-    modifier = modifier,
+    modifier = modifier.pieRotation(rotation),
   )
 }
 
--- a/vico/compose/src/androidHostTest/kotlin/com/patrykandpatrick/vico/compose/pie/PieRotationStateTest.kt
+++ b/vico/compose/src/androidHostTest/kotlin/com/patrykandpatrick/vico/compose/pie/PieRotationStateTest.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2026 by Patryk Goworowski and Patrick Michalik.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.patrykandpatrick.vico.compose.pie
+
+import androidx.compose.runtime.saveable.SaverScope
+import androidx.compose.ui.geometry.Offset
+import kotlin.test.Test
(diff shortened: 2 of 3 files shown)


adrcotfas:
@claude @patrickmichalik That's too bad. Your other charts provide a scroll option. Why wouldn't the PieChart have something equivalent built-in? If I redo with a `PieChartHost` version, would that be accepted?

adrcotfas:
@claude Thanks for the detailed feedback. Here is a sketch of the state API, following the `VicoScrollState` / `CartesianChartHost` pattern you mentioned.
``` kotlin
public class VicoPieRotationState internal constructor(angle: Float) {
  private val animatable = Animatable(angle)
  internal var rotationEnabled by mutableStateOf(true) // synced via SideEffect, like scrollEnabled
  public val angle: Float get() = animatable.value
  public suspend fun rotate(angle: Float)
  public suspend fun animateRotate(angle: Float, animationSpec: AnimationSpec<Float> = spring())
  public suspend fun stop() // cancels an in-progress drag/fling
}

@Composable
public fun rememberVicoPieRotationState(
  rotationEnabled: Boolean = false,
  initialRotation: Float = -90f,
): VicoPieRotationState

@Composable
public fun PieChartHost(
  chart: PieChart,
  modelProducer: PieChartModelProducer,
  modifier: Modifier = Modifier,
  rotationState: VicoPieRotationState = rememberVicoPieRotationState(initialRotation = chart.startAngle),
  // ...other params unchanged
)
```

Main points:

1. One source of truth. `PieChartHost` reads `rotationState.angle` each frame and passes it to `chart.draw()` directly, not through `PieChart.copy()`, so rotating does not recreate the chart config every frame. `PieChart.startAngle` becomes deprecated for host usage, but stays valid for direct `draw()/renderToImageBitmap` calls without a host. `rotationState`'s default `initialRotation` comes from `chart.startAngle`, so nothing changes for callers who touch neither.
2. Disabled by default, as you suggested. A caller who wants a fixed custom orientation with no interaction just sets `initialRotation`, and leaves `otationEnabled = false`.
3. Gesture arbitration. `Modifier.scrollable` does not fit here, since a rotation drag's useful part is tangential to the touch point, not a fixed-axis value. Plan is to use `Modifier.draggable2D` instead, which uses the same dispatch pipeline as scrollable and gets normal `NestedScrollConnection` arbitration for free. The delta gets split into a tangential part (drives rotation) and a radial part (left for the parent), instead of consuming every event like the old PR did. That should fix the "steals gesture from parent scroll" issue.
4. Bounded growth. Normalize angle % 360 only when the animation is at rest, never mid gesture, so it can spin freely without the float growing forever.
5. Cancellation. A public `stop()` wraps `Animatable.stop()` for programmatic cancel. Gesture level cancellation (second pointer, parent nested scroll taking over) comes for free from `draggable2D`.

If this direction looks right to you, I would like to implement it and open a new PR.

adrcotfas:
@claude Thanks for the correction on `draggable2D` and `scrollable2D`. I checked how `scrollable` itself works with parents, and I think the per-pixel split from my sketch was anyway a wrong model.

**Why the split does not work.** Pointer consumption in Compose is all or nothing per event (`PointerInputChange.consume()`, partial consumption was removed). Even `scrollable` does not split events. After it wins touch slop, it consumes everything, and talks to parents only through nested scroll (pre-scroll, leftover delta at bounds, fling leftover). Rotation has no bounds, so there is never leftover delta to pass. Also rotate plus scroll from one finger at the same time would feel broken.

**Proposal: decide once, at touch slop.** Custom `pointerInput` in `PieChartHost`, similar to how orientation locking decides between horizontal and vertical scrollables:

1. `awaitFirstDown` without consuming. Stop a running fling here, so a new gesture takes over.
2. `awaitTouchSlopOrCancellation`. At slop, split the drag vector relative to the pie center into tangential and radial parts. Down position must be inside the pie.
3. Tangential part wins: take the gesture, consume all next events, rotate until pointer up. Radial part wins: consume nothing, parent scroll takes it as usual.
4. During drag, rotation delta is the real angle change of the pointer around the center (atan2), not a projection, so it stays correct near the center. Applied with `Animatable.snapTo`.
5. On release, angular velocity from `VelocityTracker` (tangential velocity divided by radius), then `Animatable.animateDecay` with `exponentialDecay`. After decay settles, normalize the angle mod 360.
6. If slop or drag gets cancelled, no fling.

**Nested scroll.** I think the first version does not need `NestedScrollDispatcher` at all. Rotation never has unconsumed delta, and radial gestures reach the parent already at the pointer level (step 3). A dispatcher would only help parents that want to observe rotation deltas, which does not seem useful. It can be added later, all of this is internal.

Rest stays as sketched, without `rotate()`, `animateRotate()` and `stop()`, per your earlier comment. Does this sound right?

PR #1548, opened 2026-07-10
fix(PieChart): give slice spacing a constant width

adrcotfas:
image

fix(PieChart): give slice spacing a constant width

Slice spacing now always produces a constant-width gap, matching the
intended behavior for both pies and donuts, instead of the gap widening
toward the outer edge:

- Donuts trim each slice edge by asin((spacing / 2) / radius) at the outer
  and inner radius, so the gap width no longer narrows toward the hole.
- Holeless pies keep a constant-angle trim, but the closing point is now
  offset along the slice's bisector (`centerOffset`) rather than left at
  the true center, so the gap doesn't collapse to a point.

`createSlicePath`'s `centerOffset` now only moves that closing point,
not the whole `circleBounds`/arc. Translating the full bounds could push
narrow slices outside the circle, causing them to visibly wander.

Adds a `BasicDonutChart` sample.


adrcotfas:
@claude done

Gowsky:
Hello, @adrcotfas. Thank you for your contribution. Please ensure the code is formatted with ktfmt using `google-style`, as per the [contribution guidelines](https://guide.vico.patrykandpatrick.com/contributing#pull-requests).

adrcotfas:
Hey @Gowsky, it's done. Can you please check https://github.com/patrykandpatrick/vico/pull/1557 too?  

BTW, you're using an older version of ktfmt on CI and it evaluates differently from the latest one I used locally.

Gowsky:
@adrcotfas We'll look into #1557 and the ktfmt version soon. Thanks again!

PR #1554, opened 2026-07-14
Fix two scroll-position regressions in VicoScrollState

lewan42:
## Summary                                                                                                                                                                         
                                                                                                                                                                                     
  Two independent scroll-correctness fixes in `VicoScrollState`, one per commit. Follow-up to #1539 (visible x range preservation).                                                                                                                                 
                                                                                                                                                                                     
  ## 1. Scroll anchor drifts when data is trimmed on the start side                                                                                                                  
                                                                                                                                                                                     
  When data is trimmed on the start side while the scroll is pinned to the end, the viewport rolls toward the start by the size  of the trim instead of staying put.                                                                                                                                                                               
                                                                                                                                                                                     
  **Cause:** In `update`, the scroll `value` used to compute the anchor's start-edge x coordinate is read *after* `maxValue` is updated. The `maxValue` setter clamps `value` to the new maximum, so when the content shrinks on the start side (the  max scroll distance drops below the current, end-pinned `value`), the clamp shifts `value` back *before* the repositioning runs. The repositioning then computes `startEdgeX` from the already-clamped value and offsets the viewport by the trimmed amount.                                                                                                                                                    
                                                                                                                                                                                     
  **Fix:** Capture the pre-clamp value and use it when computing `startEdgeX`. RTL handling (`layoutDirectionMultiplier`) is preserved.
  
  ## 2. Stale zoom scroll compensation applied after a scroll                                                                                                                        
                                                                                                                                                                                     
  After a zoom, the viewport can jump back to its zoom-time position if a pan or fling starts without the scroll ever becoming idle (e.g. a pinch that transitions straight into a pan).                                                                                                                                                  
                                                                                                                                                                                     
  **Cause:** `scroll(scroll, maxScroll)` receives the zoom's scroll compensation — an absolute target computed from the scroll value and content width at the time of the zoom event — and waits for the scroll to become idle (`isScrollInProgress.first { !it }`) before applying it. If the scroll never goes idle, both the target and `maxScroll` are stale by the time it runs, so applying them rolls the viewport back to the zoom-time position.                                                                                                                            
                                                                                                                                                                                     
  **Fix:** Skip the compensation when a scroll is already in progress. The zoom handler emits a fresh compensation on every zoom event, so dropping a stale one is safe.                                                                                                                                                                           
                                                                                                                                                                                     
  ## Testing                                                                                                                                                                         
                                                                                                                                                                                     
  Added a `VicoScrollStateTest` case for each fix. Both fail without their respective fix (the viewport rolls left / the call times out) and pass with it.

<!-- codesmith:footer -->
---
<a href="https://app.blacksmith.sh/patrykandpatrick/codesmith/vico/pr/1554"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg"><img alt="View with Codesmith" src="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"></picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786608468&installation_id=136960981&pr_number=1554&repository=patrykandpatrick%2Fvico&return_to=https%3A%2F%2Fgithub.com%2Fpatrykandpatrick%2Fvico%2Fpull%2F1554&signature=0a21ffaa946ee9bbd4dede45c0ab249757b43134ec5d239f86a48cf0a913b591"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg"><img alt="Autofix with Codesmith" src="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"></picture></a>

<!-- codesmith:autofix:disabled -->
<!-- /codesmith:footer -->

### DIFF
--- a/vico/compose/src/androidHostTest/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollStateTest.kt
+++ b/vico/compose/src/androidHostTest/kotlin/com/patrykandpatrick/vico/compose/cartesian/VicoScrollStateTest.kt
@@ -208,6 +208,37 @@ class VicoScrollStateTest {
     assertEquals(-20f, sut.value)
   }
 
+  @Test
+  fun `When data is trimmed on the start side while pinned to the end, then update keeps the scroll pinned to the end`() =
+    runBlocking {
+      maxX = 19.0
+      xLength = 19.0
+      val bounds = Rect(0f, 0f, 100f, 100f)
+      val layerDimensions =
+        MutableCartesianLayerDimensions(
+          xSpacing = 10f,
+          scalableStartPadding = 6f,
+          unscalableStartPadding = 4f,
+        )
+      // The initial content is 100px wider than the chart, so the end-pinned scroll value is 100,
+      // matching the max scroll distance. The scroll is at the end.
+      val sut = createScrollState(initialScroll = Scroll.Absolute.End, layerDimensions = layerDimensions)
+      assertEquals(sut.maxValue, sut.value)
+
+      val visibleStart = context.getVisibleXRange(layerDimensions, bounds, sut.value).start
+
+      // Trim the five oldest points. The remaining content still overflows the chart, so the new
+      // max scroll distance is smaller than the old scroll value.
+      minX = 5.0
+      xLength = 14.0
+      sut.update(context = context, bounds = bounds, layerDimensions = layerDimensions)
+
+      // The visible x range is unchanged (the trimmed points were off-screen), and the scroll stays
+      // pinned to the end rather than rolling toward the start by the size of the trim.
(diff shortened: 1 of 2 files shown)


patrickmichalik (approved):
Thanks!

PR #1557, opened 2026-07-15
feat(PieChart): add drag-to-rotate support to PieChartHost

adrcotfas:
Follow-up to the discussion in #1547. This adds drag-to-rotate to `PieChartHost`, following the design agreed there.

## API

- New `PieRotationState` and `rememberPieRotationState(rotationEnabled = false, initialRotation = -90f)`. Rotation is disabled by default.
- `PieRotationState` exposes only a read-only `angle`. There’s no programmatic rotation API in this version, as agreed.
- Both `PieChartHost` overloads get a `rotationState` parameter.

## One source of truth

`PieChart.startAngle` is gone, along with the corresponding `rememberPieChart` and `PieChart.copy` parameters. The angle now lives solely in `PieRotationState`, and `PieChartHost` reads `rotationState.angle` each frame and passes it to `chart.draw()` directly, so there’s no `PieChart.copy()` per frame. `PieChart.draw()` takes the start angle as a parameter, defaulting to `-90f`, which is what host-less rendering (including `renderToImageBitmap`) uses.

This is a breaking change: callers that set `rememberPieChart(startAngle = …)` now pass `rememberPieRotationState(initialRotation = …)` to `PieChartHost` instead.

## Gesture handling

- Custom `pointerInput`. The tangential-or-radial decision is made once, at touch slop.
- Tangential drags are captured and all further events are consumed. Radial drags aren’t consumed at all, so the parent scroll receives them at the pointer level. Near ties are treated as radial, so ambiguous gestures go to the parent (bias factor 1.25).
- The rotation delta is the real `atan2` angle change of the pointer around the pie’s center, applied with `Animatable.snapTo`.
- On release, the angular velocity comes from `VelocityTracker` (tangential velocity divided by radius), capped at 1080°/s, then `Animatable.animateDecay` with `exponentialDecay`.
- The interactive area is the ring between the hole (or a 16 dp central dead zone, whichever is larger) and the outer radius. The pie’s geometry—center, outer radius, and hole radius—is recorded during `draw` and read back by the gesture detector, so it always matches what’s on screen.
- A new down stops an active fling. A cancelled gesture doesn’t fling: the drag follows the original pointer and is cancelled if its movement is consumed elsewhere or if another finger is still down when it lifts.
- The angle is normalized to [0, 360) on every idle path (after a fling settles, after a release with no fling, after a cancellation) and when the state is saved, never mid-gesture.
- There’s no `NestedScrollDispatcher` in this version, as agreed.

## Tests

The module has no compose-ui-test dependency, so the gesture logic is factored into small internal functions that plain unit tests can call. `PieRotationGesturesTest` covers capture vs. handoff (tangential, radial, toss-up), the hole and dead zone, angle deltas including the 180° seam, velocity conversion and capping, and drag tracking (velocity after a pause, multi-finger release, consumed movement). `PieRotationStateTest` covers no normalization mid-gesture, normalization on settle and after a fling, a fling superseded by a stop, and saving and restoring the state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/patrykandpatrick/codesmith/vico/pr/1557"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786711392&installation_id=136960981&pr_number=1557&repository=patrykandpatrick%2Fvico&return_to=https%3A%2F%2Fgithub.com%2Fpatrykandpatrick%2Fvico%2Fpull%2F1557&signature=5fe6153ff9e4704c65019d1469deb0de26f16014f43d1cdad998625e4071d06d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


patrickmichalik (approved):
Thanks!

